PR #3, opened 2023-08-07
Fix CORS allowed origin configuration for local setup with DEV_DOMAIN

johnnyomair:


.env
```diff
@@ -41,7 +41,7 @@ DAM_CDN_ORIGIN_HEADER=
 API_PORT=4000
 API_URL=http://${DEV_DOMAIN:-localhost}${WORKAROUND_DOTENV_ISSUE}:${API_PORT}/api
 API_URL_INTERNAL=http://localhost:$API_PORT/api
-CORS_ALLOWED_ORIGINS="^https://.*.dev.vivid-planet.cloud:$PROXY_PORT,^http:\/\/localhost:\d+,^http://192.168.\d+.\d+:80[0-9]{2}"
+CORS_ALLOWED_ORIGINS="^http://.*.dev.vivid-planet.cloud:\d+,^http:\/\/localhost:\d+,^http://192.168.\d+.\d+:80[0-9]{2}"
```
max-debug022, line 44: **Suggestion:**
^(https?://)?(.*\.dev\.vivid-planet\.cloud|localhost|192\.168\.\d{1,3}\.\d{1,3}:\d{2,4})$

This RegEx string would have the following advantages:
- Allow https and http (optional)
- Use of \d{1,3} for IP address segments allows up to a maximum of 3 digits
- {2,4} for port numbers supports 2-digit, 3digit and 4-digit formats
- Reduced repetition of similar patterns for better maintainability

max-debug022, line 44: **Suggestion:**
`^(https?://)?(.*\.dev\.vivid-planet\.cloud|localhost|192\.168\.\d{1,3}\.\d{1,3}:\d{2,4})$`

This RegEx string would have the following advantages:

- Allow https and http (optional)
- Use of \d{1,3} for IP address segments allows up to a maximum of 3 digits
- {2,4} for port numbers supports 2-digit, 3digit and 4-digit formats
- Reduced repetition of similar patterns for better maintainability

In addition to that, `corsAllowedOrigins: envVars.CORS_ALLOWED_ORIGINS.split(","),` in the `config.ts` file would not be necessary since the NestJS [CORS-Factory](https://docs.nestjs.com/security/cors) allows RegEx (https://github.com/expressjs/cors#configuration-options). That means that:
`    app.enableCors({
        origin: config.corsAllowedOrigins.map((val: string) => new RegExp(val)),` in the main.ts would also be redundant.

johnnyomair: I've improved the CORS RegExp handling:

- Used your suggested Regex, but removed https support (will never need that locally)
- Directly create the RegExp in config.ts

nsams:
As this is now no array anymore it should be named singular

Is it a good idea to have no array anymore?

johnnyomair:
> As this is now no array anymore it should be named singular

But there are still multiple origins, they're just treated in one regex. And the header is also plural

> Is it a good idea to have no array anymore?

I'm not sure. Do you think three regexes are more readable? Should we change it back, but still keep @max-debug022's optimizations?

nsams:
The nest setting is singular....

Imho we dont need array support

johnnyomair:
> The nest setting is singular....

You're right, and the header is also singular (I was looking at the Nest JSDoc, which states it plural 😅). I've changed it.

nsams (approved):

dkarnutsch (approved):

PR #4, opened 2023-08-07
Fix internationalization workflow

max-debug022:
This PR adds multilanguage support for the comet-starter.


site/src/pages/_app.page.tsx
```diff
@@ -41,7 +42,7 @@ export default function App({ Component, pageProps }: AppProps): JSX.Element {
         // for a complete strategy to couple next with react-intl
         // defaultLocale prevents missing message warning for locale defined in code,
         // see https://github.com/formatjs/formatjs/issues/251
-        <IntlProvider locale="de" defaultLocale="de">
+        <IntlProvider locale="en" defaultLocale="en" messages={getMessages()}>
```
johnnyomair, line 45: Please use the locale information form Next i18n routing for this. In a customer project we solved this using `App.getInitialProps` and a dynamic import. Should we do this here as well? We should talk about it in person

admin/.gitignore
```diff
@@ -3,9 +3,7 @@
 .env.secrets
 /node_modules
 /build
-/lang
-/lang-extracted/
```
dkarnutsch, line 7: You probably still need `lang-extracted` because otherwise `npm run intl:extract` will cause git changes

site/src/pages/_app.page.tsx
```diff
@@ -69,3 +77,13 @@ export default function App({ Component, pageProps }: AppProps): JSX.Element {
         </IntlProvider>
     );
 }
+
+const getInitialProps: typeof App.getInitialProps = async (appContext) => {
+    const locale = appContext.router.locale ?? defaultLanguage;
+
+    const [appProps, messages] = await Promise.all([App.getInitialProps(appContext), getMessages()[locale]]);
```
johnnyomair, line 84: Please use a dynamic import for the JSON language files. Otherwise all translations are in the site's HTML, for every page, increasing the download size

site/tsconfig.json
```diff
@@ -25,6 +25,7 @@
         },
         "incremental": true
     },
+    "resolveJsonModule": true,
```
johnnyomair, line 28:
```suggestion
```

This is not needed anymore.

johnnyomair (approved):

dkarnutsch (approved):

PR #8, opened 2023-08-11
Add `@comet/create-app`

max-debug022:
# Create Comet App :comet: 

Example Terminal output:

Screenshot 2023-07-28 at 12 40 16
Screenshot 2023-07-28 at 12 42 16
Screenshot 2023-07-18 at 12 22 07
Screenshot 2023-07-18 at 12 22 19

# **README**
# @comet/create-app

This is a command-line tool to create a new Comet app. It sets up a project with the necessary configuration and files
to get started quickly.

## Usage

To create a new Comet app, run the following command:

```
npx @comet/create-app
```

### Arguments

The following arguments can be passed to customize the project setup:

- `project-name` (required): Specifies the name of the project. It will be used as the directory name for the project.
- `-ni` or `--no-install`: Skips the installation of dependencies.
- `-v` or `--verbose`: Enables extra console logs for verbose output.
- `-V` or `--version`: Outputs the version number.
- `-h` or `--help`: Display help for the command.

Example usage with arguments:

```
npx @comet/create-app my-project -v
```

Example output:
```
Creating a new Comet app in /Users/maximilianpelka/Documents/Development

Cloning into './my-project'
Cloned git repository.

Successfully replaced content in 15 files

Installing dependencies...

Installation successful

Success! Created 'test-project' at '/<installation-path>/test-project'

```

This command will create a new Comet app with the name "my-project" and enable verbose output in the console logs.

### Development

To test the script locally, run the following commands:

1. Start the development process:
    ```bash
   npm start
    ```
2. Run the script:
    ```bash
    node ./bin/index.js
    ```

Testing a close to production usage of the tool can be done by calling `npm link` in the create-app/ directory. Then `npx @comet/create-app` can be used.


create-comet-app/src/util/createWorkingDirectoryCopy.ts
```diff
@@ -0,0 +1,24 @@
...
+            }
+            if (stdout && verbose) {
+                console.log(kleur.grey(stdout));
+            }
+            process.chdir(`./${projectName}`);
+            resolve();
+        });
+    }).catch(() => {
+        console.log(kleur.bgRed(`Error while cloning working directory to ${projectName}`));
+        process.exit(1);
```
nsams, line 22: Please don't do process.exit somewhere in business code

And this whole Promise construct (that you catch in the same function) is a bit strange

create-comet-app/src/util/removeShowcaseContent.ts
```diff
@@ -0,0 +1,83 @@
...
+            content.replacements.map((stringToRemove) => {
+                fileContent = fileContent.replaceAll(stringToRemove, "");
+            });
+
+            try {
+                eslint
+                    .lintText(fileContent, {
+                        filePath: content.file,
+                    })
+                    .then((lintedFileContent) => {
```
nsams, line 54: why not await?

create-comet-app/src/util/replacePlaceholder.ts
```diff
@@ -0,0 +1,36 @@
...
+import { glob } from "glob";
+import kleur from "kleur";
+
+export function replacePlaceholder(projectName: string, verbose: boolean): void {
+    const files: string[] = glob.sync([`./**`, `./.*`, `./.**/**`, `./**/.**/**`], {
+        maxDepth: 5,
+        ignore: `./.git/**`,
+    });
+    const placeholder = /[Ss]tarter/;
+    const value = projectName;
```
nsams, line 11: projectName is such a good variable name, why do you use not-so-much meaning name value?

create-comet-app/src/createCometApp.ts
```diff
@@ -0,0 +1,83 @@
...
+            if (!isValidProjectName(projectName)) {
+                console.log(kleur.bgRed("Please provide a valid project name."));
+            }
+            createCometApp({ projectName, showcaseContent, verbose });
+        });
+    program.parse();
+})();
+
+async function createCometApp(projectConfiguration: ProjectConfiguration) {
+    console.log(kleur.white(`Creating a new comet app in `) + kleur.yellow(`${process.cwd()}\n`));
```
johnnyomair, line 49:
```suggestion
    console.log(kleur.white(`Creating a new Comet app in `) + kleur.yellow(`${process.cwd()}\n`));
```

create-comet-app/src/util/validateNodeVersion.ts
```diff
@@ -0,0 +1,8 @@
+function isNodeVersionGreaterThan18(): boolean {
+    const [major, minor] = process.versions.node.split(".").map(Number);
+    return major > 18 || (major === 18 && minor >= 0);
```
johnnyomair, line 3:
```suggestion
    return major >= 18;
```

johnnyomair (commented):
Please move the script into a `create-app` folder, as we don't call it create-comet-app

dkarnutsch:
From testing:
- .github/workflows should be removed once set up

create-comet-app/src/util/validateNodeVersion.ts
```diff
@@ -0,0 +1,8 @@
+function isNodeVersionGreaterThan18(): boolean {
+    const [major, minor] = process.versions.node.split(".").map(Number);
+    return major >= 18;
+}
```
dkarnutsch, line 4: this constraint should be specified in package.json

with https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict this might not be necessary at all

create-comet-app/src/util/createInitialGitCommit.ts
```diff
@@ -0,0 +1,42 @@
+import { exec } from "child_process";
+import kleur from "kleur";
+
+export async function createInitialGitCommit() {
```
dkarnutsch, line 4: might use https://github.com/steveukx/git-js instead of calling git on your own

dkarnutsch (changes requested):

create-comet-app/src/util/createInitialGitCommit.ts
nsams: I wouldn't change that. No real benefit for those very basic actions imho

create-comet-app/README.md
```diff
@@ -0,0 +1,34 @@
...
+npx @comet/create-app
+```
+
+### Arguments
+
+The following arguments can be passed to customize the project setup:
+
+- `project-name` (required*): Specifies the name of the project. It will be used as the directory name for the project.
+- `--no-showcase` (required*): Disables the addition of showcase content in the project.
+- `--showcase` (required*): Adds showcase content to the project.
```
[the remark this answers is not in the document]
max-debug022: Thank you, I forgot to change that since the user is not prompted to enter it if not provided.

create-comet-app/src/util/validateNodeVersion.ts
max-debug022: Doesnt [engine-strict](https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict) only prohibit installing? This validation is necessary for the replaceAll function.

.prettierignore
```diff
@@ -4,4 +4,5 @@ api/
 site/
 schema.graphql
 api/uploads/
+create-comet-app/
```
nsams, line 7: you renamed that to create-app; but why do you have it in here at all?

create-app/README.md
```diff
@@ -0,0 +1,34 @@
...
+The following arguments can be passed to customize the project setup:
+
+- `project-name` (required*): Specifies the name of the project. It will be used as the directory name for the project.
+- `--no-showcase`: Disables the addition of showcase content in the project.
+- `--showcase`: Adds showcase content to the project.
+- `-v` or `--verbose`: Enables extra console logs for verbose output.
+- `-V` or `--version`: Outputs the version number.
+- `-h` or `--help`: Display help for the command.
+
+*If you choose not to provide a required argument, you will be prompted to enter it during the setup process.
```
nsams, line 25: remove that outdated line

create-app/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@comet/create-app",
+  "version": "1.0.0",
+  "bin": {
```
nsams, line 4: still 2 spaces, please change to 4

nsams (approved):

.prettierignore
```diff
@@ -3,5 +3,5 @@ admin/
 api/
 site/
 schema.graphql
-api/uploads/
```
johnnyomair, line 6: This line should not be deleted, otherwise header files (which are JSON) will be checked using Prettier on commit, which is not necessary

create-app/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+    "extends": ["@comet/eslint-config/nestjs"]
```
johnnyomair, line 2: We could add a separate `@comet/eslint-config/node` in the future

create-app/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+- `project-name` (required): Specifies the name of the project. It will be used as the directory name for the project.
+- `-v` or `--verbose`: Enables extra console logs for verbose output.
+- `-V` or `--version`: Outputs the version number.
+- `-h` or `--help`: Display help for the command.
+
+Example usage with arguments:
+
+```
+npx @comet/create-app my-project -v
```
johnnyomair, line 26: I think we should show an example output here

create-app/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+### Arguments
+
+The following arguments can be passed to customize the project setup:
+
+- `project-name` (required): Specifies the name of the project. It will be used as the directory name for the project.
+- `-v` or `--verbose`: Enables extra console logs for verbose output.
+- `-V` or `--version`: Outputs the version number.
+- `-h` or `--help`: Display help for the command.
```
johnnyomair, line 21: Do you plan to add the `--no-install, --ni` (as discussed internally) in this PR?

create-app/src/createCometApp.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function isValidProjectName(value: string): boolean {
+    const allowedFormat = /^[A-Za-z0-9][A-Za-z0-9-]*$/;
+    return allowedFormat.test(value);
+}
+
+void (async () => {
+    const program = new Command();
+    if (!validateNodeVersion()) {
```
johnnyomair, line 26: This function should be called `isValidNodeVersion()` IMO, as the actual "validation" (= the early return) happens outside the function.

create-app/src/createCometApp.ts
```diff
@@ -0,0 +1,72 @@
...
+        console.log(kleur.bgRed("Invalid Node Version (your Node.js version is prior to v18)."));
+        return;
+    }
+
+    program.name(name).description("CLI to create a comet app").version(version);
```
johnnyomair, line 31:
```suggestion
    program.name(name).description("CLI to create a Comet app").version(version);
```

create-app/src/util/replacePlaceholder.ts
```diff
@@ -0,0 +1,35 @@
...
+            if (placeholder.test(contents)) {
+                if (verbose) {
+                    console.log(kleur.grey(`Info: Replaced content in ${file}`));
+                }
+                fs.writeFileSync(file, contents.replaceAll(placeholder, projectName));
+                changedFiles++;
+            }
+        } catch (e) {
+            console.log(e);
+            console.log(kleur.yellow(`Could not replace Placeholder in ${file}`));
```
johnnyomair, line 29:
```suggestion
            console.log(kleur.yellow(`Could not replace placeholder in ${file}`));
```

.prettierignore
max-debug022: but isnt this file already covered in api/.prettierignore?

max-debug022: Shouldn't this already be covered by `api/` in line 3?

johnnyomair: Yes, but when using Prettier directly in the api/ folder, files in the uploads/ folder get formatted. Could you please add it to api/.prettierignore instead?

johnnyomair (approved):
Please wait for all reviewers to approve at least once. Then I think we should squash-merge this PR 🙂

nsams (dismissed):

johnnyomair (dismissed):

dkarnutsch (dismissed):

.prettierignore
```diff
@@ -3,5 +3,5 @@ admin/
...
+create-app/README.md
```
nsams, line 6: - if you place it here in the root it will be part of every app
- why prettierignore readme at all?

create-app/package.json
```diff
@@ -0,0 +1,47 @@
...
+        "start": "tsc --watch --project tsconfig.json"
+    },
+    "dependencies": {
+        "eslint": "^8.0.0",
+        "commander": "^11.0.0",
+        "glob": "^10.3.3",
+        "kleur": "^4.1.5",
+        "nanospinner": "^1.1.0",
+        "rimraf": "^5.0.1",
+        "ts-node": "^10.9.1"
```
nsams, line 29: I think you don't need ts-node as dependency, not even as dev-dependency?

create-app/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "devDependencies": {
+        "@types/prompts": "^2.4.4",
+        "@comet/eslint-config": "^4.3.0",
+        "@typescript-eslint/eslint-plugin": "^6.4.1",
+        "@typescript-eslint/parser": "^6.0.0",
+        "eslint-config-prettier": "^6.12.0",
+        "eslint-plugin-import": "^2.22.1",
+        "eslint-plugin-prettier": "^3.1.4",
+        "eslint-plugin-simple-import-sort": "^5.0.3",
+        "eslint-plugin-unused-imports": "^0.1.3",
```
nsams, line 40: please clenaup devDependencies

- remove @types/prompts
- remove all eslint packages that you don't use directly (@comet/eslint-config alone should be enough)
- eslint package is not a devDependency as you need it at runtime. what about plugins and configs?

nsams (approved):

dkarnutsch (dismissed):

fraxachun:
The no-install flag does not work, at least with my tests. It always skips this step.

fraxachun:
Please add instructions in readme on how to develop and publish this module.

johnnyomair:
> Please add instructions in readme on how to develop and publish this module.

I will add publishing in a follow-up release. My idea is to automatically publish a minor version on every change in `create-app/*`, similar to `@comet/upgrade`. Do you think this is a good idea?

max-debug022:
> The no-install flag does not work, at least with my tests. It always skips this step.
I have tested it and both `-ni` and `--no-install` worked as intended.

nsams:
did you also test without the flag?

johnnyomair (dismissed):

nsams (dismissed):

create-app/tsconfig.json
```diff
@@ -1,19 +1,24 @@
...
-        "noImplicitAny": true,
-        "strictNullChecks": true,
         "jsx": "react",
-        "noEmit": true,
-        "skipLibCheck": true,
         "baseUrl": "./",
         "paths": {
-            "@src/*": ["src/*"]
+            "@src/*": ["src/*"],
+            "@util/*": ["src/util/*"],
```
johnnyomair, line 9: We don't have this alias

nsams (approved):

create-app/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## For developers
+
+### Development
+
+To test the script locally, run the following commands:
+
+1. Start the development process:
+    ```
+    bash npm start
```
johnnyomair, line 57:
```suggestion
    ```bash
    npm start
```

Maybe I want to use zsh? 😉

create-app/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+    "compilerOptions": {
+        "target": "es2022",
+        "module": "CommonJS",
+        "jsx": "react",
+        "baseUrl": "./",
+        "rootDirs": ["src/", "./"],
```
johnnyomair, line 7:
```suggestion
```
I don't think we need this option

create-app/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+    "compilerOptions": {
+        "target": "es2022",
+        "module": "CommonJS",
+        "jsx": "react",
```
johnnyomair, line 5:
```suggestion
```

We're not using React in this tool

create-app/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+        "outDir": "lib/",
+        "noEmit": false,
+        "skipLibCheck": true,
+        "esModuleInterop": true,
+        "resolveJsonModule": true,
+        "allowSyntheticDefaultImports": true,
+        "strict": true,
+        "allowUnreachableCode": false,
+        "allowUnusedLabels": false,
+        "exactOptionalPropertyTypes": true
```
johnnyomair, line 17: It seems to me that you're randomly adding settings to TS config. Some may be good, others we may not need at all. Why not use the [Node 18 base](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#tsconfig-bases) as a start and _only_ add the settings we definitely need?

And: We should unify the TS configs across all Comet packages (out of scope) IMO.

johnnyomair:
Running npm install in the create-app folder leads to package-lock.json changes. Please fix that

johnnyomair:
Running `npm run build` adds new files that are not ignored:

image

Also, package.json and the src/ folder shouldn't be in lib/. This is presumably to a wrong configuration in TS config

johnnyomair:
The bin/index.js is not executable: `chmod +x bin/index.js`

For your info: You can test a close to production usage of the tool by calling `npm link` in the create-app/ directory. Then you may try `npx @comet/create-app`

create-app/package.json
```diff
@@ -0,0 +1,39 @@
+{
+    "name": "@comet/create-app",
+    "version": "1.0.0",
+    "bin": {
+        "@comet/create-app": "./bin/index.js"
+    },
```
johnnyomair, line 6:
```suggestion
    "bin": "./bin/index.js",
```

create-app/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "files": [
+        "bin/**/*.js",
```
johnnyomair, line 8:
```suggestion
        "bin/index.js",
```

max-debug022:
> The bin/index.js is not executable: `chmod +x bin/index.js`
> 
> For your info: You can test a close to production usage of the tool by calling `npm link` in the create-app/ directory. Then you can use `npx @comet/create-app`

Thank you for this hint!

create-app/tsconfig.json
max-debug022: Using Node18 base is a good idea. Nevertheless, I did not randomly add settings. I mostly used recommended settings as documented in https://www.typescriptlang.org/tsconfig.

create-app/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "extends": "@tsconfig/node18/tsconfig.json",
+    "compilerOptions": {
+        "baseUrl": "./",
+        "outDir": "lib/",
+        "noEmit": false,
+        "resolveJsonModule": true,
+        "allowSyntheticDefaultImports": true,
+        "allowUnreachableCode": false,
+        "allowUnusedLabels": false,
+        "exactOptionalPropertyTypes": true
```
fraxachun, line 11: > Option 'exactOptionalPropertyTypes' cannot be specified without specifying option 'strictNullChecks'.ts

create-app/src/createApp.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (!createWorkingDirectoryCopy(projectConfiguration.projectName, projectConfiguration.verbose)) {
+        return;
+    }
+    cleanupWorkingDirectory(projectConfiguration.verbose);
+    replacePlaceholder(projectConfiguration.projectName, projectConfiguration.verbose);
+    if (projectConfiguration.noInstall) {
+        console.log(kleur.white(`Skipping installation of dependencies`));
+    } else {
+        const spinner = createSpinner("Installing dependencies").start();
+        installProjectPackages();
```
fraxachun, line 58: > .git can't be found (see https://git.io/Jc3F9)
> npm ERR! code 1
> npm ERR! path /Users/franz/dev/test
> npm ERR! command failed
> npm ERR! command sh -c husky install

I think you have to execut `git init` first.

max-debug022:
> The description of the PR should be added to the Readme of the project

I dont understand what you mean. The description of this PR is based on the README. Or do you mean the images?

fraxachun:
> I dont understand what you mean. The description of this PR is based on the README. Or do you mean the images?

You're right, i mixed up the directories.

create-app/src/util/createInitialGitCommit.ts
```diff
@@ -0,0 +1,15 @@
+import { execSync } from "child_process";
+import kleur from "kleur";
+
+export function createInitialGitCommit() {
```
fraxachun, line 4: This is not called anywhere.

max-debug022: Thanks I deleted the function call by mistake

max-debug022:
@nsams @johnnyomair @fraxachun could you please review/approve this pr again?

nsams (approved):

create-app/src/util/replacePlaceholder.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    files.forEach((file) => {
+        try {
+            if (!fs.lstatSync(file).isFile()) {
+                return;
+            }
+            const contents = fs.readFileSync(file, "utf8").toString();
+
+            if (placeholder.test(contents)) {
+                if (file.endsWith("intl-update.sh")) fs.writeFileSync(file, contents.replaceAll("lang/starter-lang", `lang/${projectName}-lang`));
```
johnnyomair, line 21: IMO this if should be before the  placeholder.test if, as there's no need to check if the file contains starter (we already know that it does)

create-app/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "extends": "@tsconfig/node18/tsconfig.json",
+    "compilerOptions": {
+        "lib": ["es2022"],
```
johnnyomair, line 4:
```suggestion
```

This should not be needed, as there's already a never version in the base: https://github.com/tsconfig/bases/blob/main/bases/node18.json#L8

create-app/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+        "rootDir": "./src",
+        "baseUrl": "./",
```
johnnyomair, line 6: This config option is not recommended for Node environments: https://www.typescriptlang.org/tsconfig#baseUrl

create-app/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
johnnyomair, line 1: This should be enough for our use case

```json
{
    "extends": "@tsconfig/node18/tsconfig.json",
    "compilerOptions": {
        "rootDir": "src/",
        "outDir": "lib/"
    },
    "include": ["src/**/*"]
}
```

create-app/src/util/createWorkingDirectoryCopy.ts
```diff
@@ -0,0 +1,18 @@
+import { execSync } from "child_process";
+import kleur from "kleur";
+import process from "process";
+
+export function createWorkingDirectoryCopy(projectName: string, verbose: boolean): boolean {
+    const clone = `git clone --depth 1 git@github.com:vivid-planet/comet-starter.git ./${projectName}`;
```
johnnyomair, line 6: Please use a HTTPS clone for this

create-app/src/createApp.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { cleanupWorkingDirectory } from "./util/cleanupWorkingDirectory";
+import { createInitialGitCommit } from "./util/createInitialGitCommit";
+import { createWorkingDirectoryCopy } from "./util/createWorkingDirectoryCopy";
+import { installProjectPackages } from "./util/installProjectPackages";
+import { isValidNodeVersion } from "./util/isValidNodeVersion";
+import { replacePlaceholder } from "./util/replacePlaceholder";
+
+const version = require("../package.json").version;
+const name = require("../package.json").name;
```
johnnyomair, line 14: This should fail in lint. Please add create-app/ to the lint setup (lint-staged and ESLint)

johnnyomair: And this could also be like this `const { name, version } = require("../package.json");`

johnnyomair (changes requested):

create-app/src/util/replacePlaceholder.ts
max-debug022: We came to the conclusion that we leave it as it is.

create-app/src/createApp.ts
```diff
@@ -57,9 +57,9 @@ async function createApp(projectConfiguration: ProjectConfiguration) {
         console.log(kleur.white(`Skipping installation of dependencies`));
     } else {
         const spinner = createSpinner("Installing dependencies").start();
+        createInitialGitCommit();
```
johnnyomair, line 60: This call should be before creating the spinner. Otherwise the output looks like this:

image

johnnyomair: Also, it doesn't create a commit when using the option `--no-install`

johnnyomair (changes requested):

johnnyomair:
@max-debug022 please adapt the PR's description so it can be used as the commit message of the squashed commit?

create-app/package.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    "dependencies": {
+        "commander": "^11.0.0",
+        "eslint": "^8.0.0",
+        "glob": "^10.3.3",
+        "kleur": "^4.1.5",
+        "nanospinner": "^1.1.0",
+        "rimraf": "^5.0.1"
+    },
+    "devDependencies": {
```
johnnyomair, line 28: The package should have TypeScript v5 as dev dependency. Thanks @kaufmo for noticing!

create-app/src/createApp.ts
max-debug022: Thats true thanks

johnnyomair (dismissed):

lint-staged.config.js
```diff
@@ -5,5 +5,7 @@ module.exports = {
     "api/**/*.{ts,tsx}": () => "npm --prefix api run lint:tsc",
     "site/**/*.{ts,tsx,js,jsx,json,css,scss,md}": () => "npm --prefix site run lint:eslint",
     "site/**/*.{ts,tsx}": () => "npm --prefix site run lint:tsc",
+    "create-app/**/*.{ts,tsx,js,jsx,json,css,scss,md}": () => "npm --prefix create-app run lint:eslint",
+    "create-app/**/*.{ts,tsx}": () => "npm --prefix create-app run lint:tsc",
```
nsams, line 9: if you add those lines (are they really required?) you have to remove them in create-app again...

create-app/src/util/installProjectPackages.ts
```diff
@@ -0,0 +1,12 @@
+import { execSync } from "child_process";
+import kleur from "kleur";
+
+export function installProjectPackages(verbose: boolean): void {
+    try {
+        const stdout = execSync("./install.sh");
+        if (verbose) console.log(kleur.white(stdout.toString()));
```
nsams, line 7: for future improvements:
this now logs everything after install finished, better would be to stream output while process is running (as this one is running quite long, and itself has progress output)

lint-staged.config.js
max-debug022: No those lines are not required for the code but aren't necessary for the pre-commit hook?

fraxachun (changes requested):

create-app/src/util/installProjectPackages.ts
fraxachun: I would skip this step at all.

- It takes very long
- Outputting doesn't work (neither streamed nor after finishing)
- Interrupting leaves you with the child processes still running
- Executing install.sh is a common task for every developer

nsams: everything correct.

The agument for doing the install is that others (create-next-app, create-react-app) also do the install step.

Personally, I don't care :)

lint-staged.config.js
nsams: Of course we want to lint everything including create-app. The lint pipeline doesn't use lint-staged though (i think) - if that is correct you might not need it here - if you can accept that pre-commit hook doesn't lint create-app.

Or, you simply remove those lines during project setup

create-app/src/util/installProjectPackages.ts
fraxachun: If you don't want to delete it, use `{ stdio: "inherit" }` to stream the output. However, I don't know how to interrupt it.

nsams (dismissed):

fraxachun (approved):

dkarnutsch (approved):

create-app/src/util/installProjectPackages.ts
johnnyomair: > Interrupting leaves you with the child processes still running

IMO this should be fixed asap

nsams: the simplest fix would be to remove it

johnnyomair (approved):

create-app/src/util/installProjectPackages.ts
fraxachun: As I already stated, I'm for removing.

johnnyomair: Do we need to decide this now, or do it in a follow-up PR?

johnnyomair (changes requested):
Please remove installing the dependencies, as it doesn't work properly.

This includes
- Removing the `--no-install` option
- Changing the final output to suggest executing `./install.sh`

create-app/src/util/installProjectPackages.ts
johnnyomair: I'd rather remove it than having a solution that doesn't work properly.

nsams (dismissed):

johnnyomair (approved):

create-app/src/createApp.ts
```diff
@@ -0,0 +1,69 @@
...
+    console.log(kleur.white(`Creating a new Comet app in `) + kleur.yellow(`${process.cwd()}\n`));
+    if (!createWorkingDirectoryCopy(projectConfiguration.projectName, projectConfiguration.verbose)) {
+        return;
+    }
+    cleanupWorkingDirectory(projectConfiguration.verbose);
+    replacePlaceholder(projectConfiguration.projectName, projectConfiguration.verbose);
+    createInitialGitCommit();
+    console.log(`\n${kleur.white(`Success! Created '${projectConfiguration.projectName}' at '${process.cwd()}'.`)}`);
+    console.log(kleur.white(`Inside that directory, you can run several commands:\n`));
+    console.log(kleur.white(`nvm use\n`));
```
johnnyomair, line 55: Please add "Run ./install.sh" or similar here

johnnyomair (changes requested):

johnnyomair (approved):

PR #17, opened 2023-09-13
Fix intl update

max-debug022:


admin/intl-update.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+cd "$(dirname "$0")" || exit
+
+rm -rf ./lang/
+mkdir -p ./lang
+
+git clone https://github.com/vivid-planet/comet-starter-lang.git lang/starter-lang
```
dkarnutsch, line 8: what about comet messages?

dkarnutsch (changes requested):

.github/workflows/lint.yml
```diff
@@ -29,19 +29,26 @@ jobs:
...
           token: ${{ secrets.GITHUB_TOKEN }}
-          path: "lang/starter-lang"
+          path: "admin/lang/starter-lang"
+
+      - name: "Site: Clone starter translations"
+        uses: actions/checkout@v3
+        with:
+          repository: vivid-planet/comet-starter-lang
+          token: ${{ secrets.GITHUB_TOKEN }}
+          path: "site/lang/starter-lang"
```
thomasdax98, line 51: @dkarnutsch Is it a good idea to clone the translations here?

It looks to me like this was done in a postinstall script in the old starter. That seems simpler to me because you don't have to treat the pipeline and dev environment (./install.sh) separately.

dkarnutsch (approved):

nsams (approved):

thomasdax98 (approved):

PR #18, opened 2023-09-14
Don't disable intl messages in development mode

nsams:
I don't see any reason why this should be done


dkarnutsch (approved):
Yes, this was also the decision from one JFX

nsams:
If you change the default message of an already translated message you have a problem anyway. You won't fix it by hiding it in development...

johnnyomair:
You're right. Especially as we decided to support changing the messages of the default locale without a code update, if I remember correctly.

johnnyomair (dismissed):

chernylu (dismissed):

max-debug022 (dismissed):

dkarnutsch (dismissed):

dkarnutsch (approved):

PR #20, opened 2023-09-19
Admin: Use Vite instead of Webpack

mennoxx:
Advantages:
- Faster Build Process (tested on M2 Max)
    - production build (npm run build):
        - webpack: 79.78s
        - vite: 27.60s
- Native ES Module Support
- Fast starting dev-server with HMR

Noteworthy changes:
- static assets from public directory are automatically copied to build directory on build. They can be imported (for dev and production) with "/path-to-file". More Information: https://vitejs.dev/guide/assets.html
- To check the current environment during runtime `import.meta.MODE` should be used instead of `process.env`. More Information: https://vitejs.dev/guide/env-and-mode.html
- Changed tsconfig setting `jsx` to `react-jsx` which allows using jsx without importing react. More Information: https://dev.to/titungdup/you-no-longer-need-to-import-react-from-react-3pbj


admin/package.json
```diff
@@ -3,9 +3,10 @@
     "private": true,
     "scripts": {
         "postinstall": "cd server && npm install",
-        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -- cross-env BABEL_ENV=development webpack serve --config ./webpack.config.ts --mode development",
```
nsams, line 6: by removing `dotenv -c secrets` and switching to dotenv included in vite we loose the possibility to use the secrets env file.

Question is of course if we ever need secrets on client side.

admin/tsconfig.json
```diff
@@ -13,7 +19,12 @@
...
-        "allowSyntheticDefaultImports": true
+        "allowSyntheticDefaultImports": true,
+        "sourceMap": true,
+        "inlineSources": true,
+        "sourceRoot": "~/",
+        "types": ["vite/client"]
     },
-    "include": ["src/**/*"]
+    "include": ["src/**/*"],
+    "exclude": ["src/loader.ts"]
```
nsams, line 29: why exclude that?

admin/tsconfig.json
```diff
@@ -1,10 +1,16 @@
 {
+    "ts-node": {
```
nsams, line 2: is ts-node used in admin?

mennoxx: Yes, this setting is needed for all node scripts (like intl:update). Without it, scripts fail saying "Cannot use import statement outside a module".

admin/index.html
```diff
@@ -6,12 +6,11 @@
         <title>Starter Admin</title>
         <script type="text/javascript">
             (function() {
-<%= environmentValues.map((env) => `window.EXTERNAL__${env.key}__ = '${env.value}';`).join("\n") %>
+                <%- environmentValues %>
```
johnnyomair, line 9: Could the template be still a .ejs file? Otherwise there are syntax errors in the editor:
image

admin/vite.config.ts
```diff
@@ -0,0 +1,53 @@
...
+                    },
+                },
+            }),
+        ],
+        server: {
+            port: Number(envVars.ADMIN_PORT || 8002),
+        },
+        define: {
+            // define NODE_ENV for packages using it
+            "process.env.NODE_ENV": mode === "production" ? "'production'" : "'development'",
```
johnnyomair, line 33: If we still define it, couldn't we continue to use it? This way it would stay the same for API/Admin/Site

admin/vite.config.ts
```diff
@@ -0,0 +1,53 @@
...
+                            .map(
+                                (envVarKey) => `window.EXTERNAL__${envVarKey}__ = '${mode === "production" ? `$${envVarKey}` : envVars[envVarKey]}';`,
+                            )
+                            .join("\n"),
+                    },
+                },
+            }),
+        ],
+        server: {
+            port: Number(envVars.ADMIN_PORT || 8002),
```
johnnyomair, line 29:
```suggestion
            port: Number(envVars.ADMIN_PORT || 8000),
```
We're using 8000 in all our projects, I wouldn't change that. But why do we even need the fallback?

admin/package.json
johnnyomair: As discussed in our meeting: Please don't remove `dotenv -c secrets`, as we use it locally for our "secrets" injection mechanism

admin/tsconfig.json
nsams: hm, those node scripts don't use ts-node?

mennoxx: If they are written in Typescript, they need to be run with ts-node. Intl-update.ts is called with ts-node in its npm command.

admin/index.html
mennoxx: I also didn't like that and tried .ejs, but couldn't get it working yet.

admin/vite.config.ts
mennoxx: yes we could, but I would go with Vites recommendation to use import.meta.

admin/vite.config.ts
mennoxx: Could be useful if 8000 is already used by another process. This was directly taken from webpack.config.ts.

admin/tsconfig.json
nsams: since https://github.com/vivid-planet/comet-starter/pull/4 intl-update isn't typescript anymore (it's changed again in https://github.com/vivid-planet/comet-starter/pull/17, but is still a sh script)

admin/vite.config.ts
nsams: I vote for removing.

admin/vite.config.ts
nsams: either:

- we define all existing evns here
- or we use imports.meta

mixing (process.env for NODE_ENV and imports.meta for all others) is confusing

mennoxx: Also: with this definition we can only use NODE_ENV, not any other env variable.

admin/tsconfig.json
mennoxx: You're right. I retested all scripts without this ts-node config and they worked.

admin/vite.config.ts
mennoxx: ok, I removed it. Now it's just using the environment entry.

admin/package.json
```diff
@@ -74,22 +74,15 @@
...
         "@types/react-router-dom": "^5.1.7",
-        "@types/webpack-env": "^1.13.9",
-        "babel-loader": "^8.0.0",
-        "babel-preset-react-app": "^10.0.0",
+        "@vitejs/plugin-react-swc": "^3.3.2",
         "chokidar-cli": "^2.1.0",
-        "cross-env": "^7.0.3",
-        "css-loader": "^6.0.0",
-        "dotenv-cli": "^6.0.0",
+        "dotenv": "^16.3.1",
```
nsams, line 79: you are yusing dotenv-cli, not dotenv (although the binary is called dotenv)

admin/package.json
```diff
@@ -3,9 +3,10 @@
     "private": true,
     "scripts": {
         "postinstall": "cd server && npm install",
-        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -- cross-env BABEL_ENV=development webpack serve --config ./webpack.config.ts --mode development",
-        "build": "npm run intl:compile && run-p gql:types generate-block-types && cross-env BABEL_ENV=production webpack --config ./webpack.config.ts --mode production --progress",
         "serve": "cd server && npm run serve",
+        "start": "npm run intl:compile intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -- vite",
```
nsams, line 7: you run intl:compile twice

nsams (approved):

admin/src/App.tsx
```diff
@@ -3,6 +3,7 @@ import "@fontsource/roboto/400.css";
 import "@fontsource/roboto/500.css";
 import "@fontsource/roboto/700.css";
 import "material-design-icons/iconfont/material-icons.css";
+import "@src/polyfills";
```
fraxachun, line 6: Where does this come from?

admin/src/common/ContentScopeConfig.ts
```diff
@@ -0,0 +1,16 @@
+import {
```
fraxachun, line 1: I already found out why you were creating this file, but it would be good if your PR-description could briefly describe the changes and - more importantly - why you made them.

admin/vite.config.ts
```diff
@@ -0,0 +1,59 @@
...
+        define: {
+            // define NODE_ENV for packages using it
+            "process.env.NODE_ENV": mode === "production" ? "'production'" : "'development'",
+        },
+        optimizeDeps: {
+            esbuildOptions: {
+                // Node.js global to browser globalThis. https://github.com/vitejs/vite/discussions/5912
+                // "gloabl is not defined" occures dirctly after loading. used by draft-is package
+                define: {
+                    global: "globalThis",
```
fraxachun, line 46: I think this is a better method: https://github.com/vitejs/vite/discussions/7915#discussioncomment-3910940

admin/src/App.tsx
mennoxx: https://github.com/vivid-planet/comet-starter/blob/55d2ab4ac5ca2f7b6e0e992e1c892c18fb2a608e/admin/webpack.config.ts#L28

With Webpack this was chained before loader.ts in the entry point setting. I also tried this with vite-plugin-html but it didn't work.

admin/src/common/ContentScopeConfig.ts
mennoxx: You're right. I added a section with noteworthy changes.

admin/vite.config.ts
mennoxx: I couldn't get this method to work correctly unfortunately. Also for a small it fix seamed to produce a lot of overhead for the build. The built File was larger for example.

admin/tsconfig.json
mennoxx: I had to readd the ts-node commonjs config, because admin-generator uses typescript and can't be executed via npm if this is not set.

admin/src/lang.ts
```diff
@@ -16,6 +16,14 @@ const projectMessages = {
...
 
 export const getMessages = (): ResolvedIntlConfig["messages"] => {
+<<<<<<< HEAD
+    // in dev mode we use the default messages to have immediate changes
+    if (import.meta.env.MODE === "development") {
+        return {};
+    }
+
+=======
+>>>>>>> origin/main
```
nsams, line 26:
```suggestion
```

package.json
```diff
@@ -24,6 +24,7 @@
     },
     "devDependencies": {
         "@comet/dev-process-manager": "^2.0.0",
+        "@swc/plugin-emotion": "^2.5.120",
```
nsams, line 27: why add that at the root package?

mennoxx: not intentional

admin/tsconfig.json
```diff
@@ -13,7 +20,10 @@
         },
         "esModuleInterop": true,
         "resolveJsonModule": true,
-        "allowSyntheticDefaultImports": true
+        "allowSyntheticDefaultImports": true,
+        "sourceMap": true,
+        "inlineSources": true,
+        "types": ["vite/client"]
```
johnnyomair, line 26: Why do we need this in vendors.d.ts and here?

admin/vite.config.mts
```diff
@@ -0,0 +1,74 @@
...
+            port: Number(process.env.ADMIN_PORT),
+        },
+        define: {
+            // define NODE_ENV for packages using it
+            "process.env.NODE_ENV": mode === "production" ? "'production'" : "'development'",
+        },
+        optimizeDeps: {
+            esbuildOptions: {
+                // Node.js global to browser globalThis. https://github.com/vitejs/vite/discussions/5912
+                // "gloabl is not defined" occures dirctly after loading. used by draft-is package
```
johnnyomair, line 59:
```suggestion
                // "global is not defined" occurs directly after loading. Used by draft-is package
```

admin/tsconfig.json
mennoxx: If I recall correctly, only after adding this here, I could see the types for import.meta in my IDE.
Which way do you prefer? Then I will test it again with one declaration.

johnnyomair: I believe vendors.d.ts is better due to the issues described by @thomasdax98 here: https://github.com/vivid-planet/comet/pull/1840#discussion_r1526500119

mennoxx: Interesting. And I see this was already cross referenced.
I tested this again, and my IDE now also shows the types with the `vendors.d.ts` entry only.

nsams (dismissed):

johnnyomair:
@mennoxx please fix conflicts, thanks!

johnnyomair (approved):

thomasdax98 (approved):

PR #31, opened 2023-10-20
Display linebreaks in RTE correctly

juliawegmayr:
Line breaks in the RTE block disappeared. To correctly display the text, a TextWrapper was implemented to set paragraph spaces.

### DIFF
--- a/site/src/common/blocks/RichTextBlock.tsx
+++ b/site/src/common/blocks/RichTextBlock.tsx
@@ -25,61 +25,61 @@ const defaultRenderers: Renderers = {
      */
     blocks: {
         // Paragraph
-        unstyled: (children, { keys }) => children.map((child, idx) => <StyledComponent key={keys[idx]}>{child}</StyledComponent>),
+        unstyled: (children, { keys }) => children.map((child, idx) => <Text key={keys[idx]}>{child}</Text>),
         // Headlines
         "header-one": (children, { keys }) =>
             children.map((child, idx) => (
-                <StyledComponent as="h1" key={keys[idx]}>
+                <Text as="h1" key={keys[idx]}>
                     {child}
-                </StyledComponent>
+                </Text>
             )),
         "header-two": (children, { keys }) =>
             children.map((child, idx) => (
-                <StyledComponent as="h2" key={keys[idx]}>
+                <Text as="h2" key={keys[idx]}>
                     {child}
-                </StyledComponent>
+                </Text>
             )),
         "header-three": (children, { keys }) =>
             children.map((child, idx) => (
-                <StyledComponent as="h3" key={keys[idx]}>
+                <Text as="h3" key={keys[idx]}>
                     {child}
-                </StyledComponent>
+                </Text>
             )),
         "header-four": (children, { keys }) =>
             children.map((child, idx) => (
-                <StyledComponent as="h4" key={keys[idx]}>
+                <Text as="h4" key={keys[idx]}>
                     {child}
-                </StyledComponent>
+                </Text>
             )),
         "header-five": (children, { keys }) =>
             children.map((child, idx) => (
-                <StyledComponent as="h5" key={keys[idx]}>
+                <Text as="h5" key={keys[idx]}>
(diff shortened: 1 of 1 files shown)


nsams:
I'm not sure this is the best solution. Would replacing \n with `<br />` be an option?

johnnyomair:
I believe this should be done in the renderers, not with a wrapper.

> I'm not sure this is the best solution. Would replacing \n with `<br />` be an option?

We'd need `dangerouslySetInnerHTML` for this I suppose

nsams:
if done properly (ideally in our library) I wouldn't exclude dangerouslySetInnerHTML. If white-space: pre-line; is the best solution us that...

juliawegmayr:
I now applied the styling on the renderer. But I guess, I still need the styled-component here, as I don't know how to apply the pseudo:class otherwise.

thomasdax98:
@nsams @johnnyomair @jamesricky Can you comment on this again?

jamesricky:
> if done properly (ideally in our library) I wouldn't exclude dangerouslySetInnerHTML. If white-space: pre-line; is the best solution, use that...

I think `white-space: pre-line;` is fine 🤷🏼

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -24,7 +25,12 @@ const defaultRenderers: Renderers = {
      */
     blocks: {
         // Paragraph
-        unstyled: (children, { keys }) => children.map((child, idx) => <p key={keys[idx]}>{child}</p>),
+        unstyled: (children, { keys }) =>
+            children.map((child, idx) => (
+                <TextWrapper key={keys[idx]}>
```
johnnyomair, line 30: Please don't add an additional div, replace the `p` tag with a `Paragraph` or `Unstyled` styled-component (`styled.p`)

johnnyomair (changes requested):

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -84,3 +90,16 @@ export const RichTextBlock = withPreview(
...
+    white-space: pre-line;
+
+    &:empty {
+        margin-bottom: 0;
+
+        :before {
+            white-space: pre;
+            content: " ";
+        }
+    }
```
johnnyomair, line 104: Why do you need this?

juliawegmayr: It's necessary if an empty paragraph is used to set spaces between paragraphs.

nsams:
Soft, I suppose

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -84,3 +119,16 @@ export const RichTextBlock = withPreview(
 export function hasDraftContent(draftContent: RawDraftContentState): boolean {
     return !(draftContent.blocks.length == 1 && draftContent.blocks[0].text === "");
 }
+
+const StyledComponent = styled.p`
```
johnnyomair, line 123: I'm not particularly fond of that naming. Maybe change it to `Text` instead?

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -84,3 +119,15 @@ export const RichTextBlock = withPreview(
 export function hasDraftContent(draftContent: RawDraftContentState): boolean {
     return !(draftContent.blocks.length == 1 && draftContent.blocks[0].text === "");
 }
+
+const Text = styled.p`
+    white-space: pre-line;
+
+    // Workaround empty paragraphs used as spacing
```
johnnyomair, line 126:
```suggestion
    // Workaround when empty paragraphs are used as "spacing" in content
```

johnnyomair (changes requested):

johnnyomair (approved):

johnnyomair:
@nsams @jamesricky please review. Thanks!

nsams (approved):

jamesricky (approved):

PR #38, opened 2023-11-16
Add missing intl:update calls

johnnyomair:


nsams:
`intl-update.sh` is executed by install.sh and that should be enough. I don't want to slow down startup during development....

johnnyomair:
Translations are also updated in CI by pulling the lang repository. In our customer project we will add `intl-update.sh` calls to the pipeline. So we don't need this change.

PR #42, opened 2023-11-26
Add missing peer dep axios.

dkarnutsch:
Axios is specified as peer dep in cms-admin (https://github.com/vivid-planet/comet/blob/main/packages/admin/cms-admin/package.json#L152C25-L152C25)

### DIFF
--- a/admin/package-lock.json
+++ b/admin/package-lock.json
@@ -24,6 +24,7 @@
                 "@mui/styles": "^5.8.6",
                 "@mui/system": "^5.8.6",
                 "@mui/x-data-grid": "^5.17.26",
+                "axios": "^0.21.0",
                 "date-fns": "^2.28.0",
                 "exceljs": "^3.4.0",
                 "file-saver": "^2.0.2",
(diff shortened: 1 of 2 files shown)


johnnyomair:
Two thoughts:

1. NPM installs peer dependencies automatically since v7, why do we need to add it? I doubt we added all other peer dependencies
2. Why is Axios a peer dependency at all (probably my fault)? Maybe we could convert it to a normal dependency instead

nsams:
we use axios for uploading dam files only, that could be easily replaced by fetch

> NPM installs peer dependencies automatically since v7, why do we need to add it? I doubt we added all other peer dependencies

I prefer doing this explicitly

> Why is Axios a peer dependency at all (probably my fault)? Maybe we could convert it to a normal dependency instead

in it's current form of usage (where createHttpClient is in @comet) we usually don't have an axios import in the application. But having such a factory method that is only configurable by the apiUrl is weird anyway.

And the name httpClient (and not axios) is even more weird.

dkarnutsch:
> I prefer doing this explicitly

same


> in it's current form of usage (where createHttpClient is in @comet) we usually don't have an axios import in the application.

Yes. I found the problem because I needed to extend createHttpClient (PR in comet will follow), because I needed to add another header.

nsams (approved):

thomasdax98 (approved):

PR #46, opened 2023-11-29
Copy YoutubeVideoBlock from Demo

thomasdax98:
The YoutubeVideoBlock was updated in Demo (to support `loop` and URLs in the `youtu.be` format). It was never updated in the Starter. 

I copied the block from Demo to Starter.


johnnyomair (commented):
I believe @nsams and I explicitly removed the YouTubeVideoBlock.sc.ts file when first integrating the CMS parts into the Starter (because we don't want to promote this pattern). There are no other .sc.ts files in the Starter at the moment. Should we keep it this way?

PR #50, opened 2023-12-05
Translate products page top level title

dkarnutsch:


johnnyomair (approved):

nsams (approved):

admin/src/products/ProductsPage.tsx
```diff
@@ -9,7 +9,7 @@ export function ProductsPage(): React.ReactElement {
     const intl = useIntl();
 
     return (
-        <Stack topLevelTitle="Products">
+        <Stack topLevelTitle={intl.formatMessage({ id: "products.title", defaultMessage: "Products" })}>
```
thomasdax98, line 12: In our docs we write 

> All strings that should be translated must be masked using [`<FormattedMessage />`](https://formatjs.io/docs/react-intl/components#formattedmessage) or, if not possible, [`intl.formatMessage()`](https://formatjs.io/docs/react-intl/api#formatmessage).
>
> \- https://docs.comet-dxp.com/docs/internationalization/

So you should use `FormattedMessage` here

johnnyomair: Our Stack has problems with React.ReactNode that are yet to be fixed, so I believe using string here seems more appropriate

thomasdax98: Okay, I wasn't aware of that

thomasdax98 (approved):

PR #66, opened 2023-12-21
@comet/create-app: Automatic publishing

johnnyomair:


thomasdax98 (approved):

.github/workflows/publish-create-app.yml
```diff
@@ -0,0 +1,56 @@
+name: Publish @comet/create-app
+
+on:
+  pull_request:
+    types:
+      - closed
```
nsams, line 6: Is this correct?

nsams (approved):

.github/workflows/publish-create-app.yml
johnnyomair: Yes, closed is used for both merged and closed PRs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

johnnyomair: That's why I added the check if it is merged: https://github.com/vivid-planet/comet-starter/pull/66/files#diff-6519865ce3a86143d0919d3d3f1a5d1e17830094d887e57bb277d1555523a4c2R15

PR #97, opened 2024-01-04
Increase JSON request limit to avoid 413 errors

thomasdax98:
Increase JSON request limit to 1mb

We had to do this in multiple client projects already because we ran into 413 errors when pages got too big:

Bildschirmfoto 2024-01-04 um 12 11 00

### DIFF
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -50,7 +50,7 @@ async function bootstrap(): Promise<void> {
             contentSecurityPolicy: false, // configure this when API returns HTML
         }),
     );
-    app.use(json({ limit: "1mb" }));
+    app.use(json({ limit: "1mb" })); // increase default limit of 100kb for saving large pages
     app.use(compression());
     app.use(cookieParser());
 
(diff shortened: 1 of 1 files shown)


api/src/main.ts
```diff
@@ -48,6 +50,7 @@ async function bootstrap(): Promise<void> {
             contentSecurityPolicy: false, // configure this when API returns HTML
         }),
     );
+    app.use(json({ limit: "1mb" }));
```
dkarnutsch, line 53: I would propose to add a comment here to clarify why we need this

thomasdax98:
```suggestion
    // The default limit is 100kb
    // savePage() requests send the whole content of a page in one request
    // For large pages, this request often exceeds 100kb
    app.use(json({ limit: "1mb" }));
```

Like this?

nsams:
```suggestion
    app.use(json({ limit: "1mb" })); // increase default limit of 100kb for saving large pages
```

dkarnutsch (approved):

nsams (approved):

johnnyomair (approved):

PR #115, opened 2024-01-12
COM-243: enable typescript strict flag for all microservices

kaufmo:


site/src/pages/[[...path]].page.tsx
```diff
@@ -23,7 +24,7 @@ interface PageProps {
     documentType: string;
     id: string;
 }
-export type PageUniversalProps = PageProps & GQLPage;
+export type PageUniversalProps = PageProps & GQLPage & PropsWithLayout<GQLPageQuery>;
```
Comment (nsams) [inline site/src/pages/[[...path]].page.tsx:27]: GQLPageQuery if from the specific page-Type "Page", but this Component is responsible for all page types (though in starter no others exist)

(and it is imported from an private sibling)

why do you change that anyway?

admin/tsconfig.json
```diff
@@ -2,8 +2,8 @@
     "compilerOptions": {
         "target": "ESNext",
         "module": "CommonJS",
-        "noImplicitAny": true,
-        "strictNullChecks": true,
+        "strict": true,
+        "strictFunctionTypes": false,
```
nsams, line 6: why that?

api/tsconfig.json
```diff
@@ -14,7 +14,6 @@
         "esModuleInterop": true,
         "strict": true,
         "strictPropertyInitialization": false,
```
johnnyomair, line 16: Could we remove this?

site/src/documents/pages/Page.tsx
```diff
@@ -22,7 +22,7 @@ export const pageQuery = gql`
     }
 `;
 
-export function Page(props: PropsWithLayout<GQLPageQuery>): JSX.Element {
```
johnnyomair, line 25: Isn't GQLPageQuery the correct type?

site/src/pages/[[...path]].page.tsx
Comment (johnnyomair) [reply site/src/pages/[[...path]].page.tsx:27]: We should remove the PageUniversalProps type because it's not correct

Comment (kaufmo) [reply site/src/pages/[[...path]].page.tsx:27]: I changed it because for now it throws incompatible errors and we use it like that in many projects

admin/tsconfig.json
kaufmo: because extending an interface is no option for abstract interfaces and results in many errors. I talked with @thomasdax98 about that and we decided to ignore this rule here like we do in api for strictPropertyInitialization

api/tsconfig.json
kaufmo: yes we can but we have to rework many many classes because we do not define any initializer in the classes? for now this would mean we have to adapt 13 files with 56 initializer errors?

site/src/documents/pages/Page.tsx
kaufmo: nope because in [[...path]].page.tsx we use the generic GQLPage type and not the GQLPageQuery type which is defined here, these types are not in sync

admin/tsconfig.json
nsams: > because extending an interface is no option for abstract interfaces and results in many errors

please elaborate on that, what errors?

api/tsconfig.json
nsams: The question is not if it is something we have to adapt in starter, it's more if we want that in future projects?

Show us some examples where this error happens? And how it could be fixed?

site/src/pages/[[...path]].page.tsx
Comment (nsams) [reply site/src/pages/[[...path]].page.tsx:27]: > and we use it like that in many projects

that doesn't make it correct :)

Comment (kaufmo) [reply site/src/pages/[[...path]].page.tsx]: > that doesn't make it correct :)

I know ;) it should be resolved now

api/tsconfig.json
kaufmo: here are some examples where all properties have to be defined then:

for example here we have to define every single property in CurrentUser
```ts
// src/auth/current-user.ts
@ObjectType()
export class CurrentUser implements CurrentUserInterface {
    @Field(() => ID)
    id: string;
    @Field()
    name: string;
    @Field()
    email: string;
    @Field()
    language: string;
    @Field()
    role: string;
    @Field(() => [String], { nullable: true })
    domains: string[];
}
```

for entites without default values it has to be defined too
```ts
// src/products/entities/product.entity.ts
@ObjectType()
@Entity()
export class Product extends BaseEntity<Product, "id"> {
    [OptionalProps]?: "createdAt" | "updatedAt";

    @Field(() => ID)
    @PrimaryKey({ type: "uuid" })
    id: string = v4();

    @Field()
    @Property({ columnType: "uuid" })
    creatorId: string;

    @Field()
    @Property({
        columnType: "text",
    })
    name: string;

    @Field()
    @Property({
        columnType: "text",
    })
    description: string;

    @Field()
    @Property({
        columnType: "timestamp with time zone",
    })
    createdAt: Date = new Date();

    @Field({ nullable: true })
    @Property({
        columnType: "timestamp with time zone",
        onUpdate: () => new Date(),
    })
    updatedAt: Date = new Date();
}
```

Do you need more examples?

site/src/pages/[[...path]].page.tsx
```diff
@@ -19,11 +18,17 @@ import * as React from "react";
...
 import { GQLPagesQuery, GQLPagesQueryVariables, GQLPageTypeQuery, GQLPageTypeQueryVariables } from "./[[...path]].page.generated";
 
-interface PageProps {
-    documentType: string;
+type PageType = { documentType: "Page" } & React.ComponentProps<typeof PageTypePage>;
+export type PageUniversalProps = {
     id: string;
-}
-export type PageUniversalProps = PageProps & GQLPage;
+} & PageType;
```
Comment (nsams) [inline site/src/pages/[[...path]].page.tsx:24]: You still have the props of a concrete document type in PageUniversalProps

Comment (johnnyomair) [reply site/src/pages/[[...path]].page.tsx:24]: createGetUniversalProps returns concrete document types, so I believe this is okay?

admin/tsconfig.json
johnnyomair: The problem is with the default values for input and output in `DocumentInterface`: https://github.com/vivid-planet/comet/blob/main/packages/admin/cms-admin/src/documents/types.ts#L35-L38

This and inference cause crazy errors:

```
Type '{ Page: DocumentInterface<Pick<GQLPage, "content" | "seo">, GQLPageInput> & DependencyInterface; Link: DocumentInterface<...> & DependencyInterface; }' is not assignable to type 'Record<string, DocumentInterface<Record<string, unknown>, Record<string, unknown>>> | ((category: string) => Record<string, DocumentInterface<Record<string, unknown>, Record<...>>>)'.
  Type '{ Page: DocumentInterface<Pick<GQLPage, "content" | "seo">, GQLPageInput> & DependencyInterface; Link: DocumentInterface<...> & DependencyInterface; }' is not assignable to type 'Record<string, DocumentInterface<Record<string, unknown>, Record<string, unknown>>>'.
    Property 'Page' is incompatible with index signature.
      Type 'DocumentInterface<Pick<GQLPage, "content" | "seo">, GQLPageInput> & DependencyInterface' is not assignable to type 'DocumentInterface<Record<string, unknown>, Record<string, unknown>>'.
        Types of property 'updateMutation' are incompatible.
          Type 'TypedDocumentNode<GQLUpdatePageMutation, GQLUpdatePageMutationVariables<Pick<GQLPage, "content" | "seo">>> | undefined' is not assignable to type 'TypedDocumentNode<GQLUpdatePageMutation, GQLUpdatePageMutationVariables<Record<string, unknown>>> | undefined'.
            Type 'TypedDocumentNode<GQLUpdatePageMutation, GQLUpdatePageMutationVariables<Pick<GQLPage, "content" | "seo">>>' is not assignable to type 'TypedDocumentNode<GQLUpdatePageMutation, GQLUpdatePageMutationVariables<Record<string, unknown>>>'.
              Type 'GQLUpdatePageMutationVariables<Record<string, unknown>>' is not assignable to type 'GQLUpdatePageMutationVariables<Pick<GQLPage, "content" | "seo">>'.
                Type 'Record<string, unknown>' is missing the following properties from type 'Pick<GQLPage, "content" | "seo">': content, seo
```

api/tsconfig.json
johnnyomair: If we enable strictPropertyInitialization, we also need to add ! to all block fields, and https://github.com/vivid-planet/comet-starter/blob/main/api/src/config/environment-variables.ts. I'm not sure if we want this

johnnyomair:
Superseded by #314.

PR #120, opened 2024-01-24
Enhance README

dkarnutsch:
We lose the project name in the README but I don't think it has a real value.


create-app/src/util/cleanupReadme.ts
```diff
@@ -0,0 +1,26 @@
+import * as fs from "fs";
+
+const readmePath = "README.md";
+
+export function cleanupReadme() {
```
johnnyomair, line 5: You could also use `findIndex` to  find the line with the comment and remove all lines before it using `slice` or `splice`. Maybe the code would be easier to understand?

nsams: or `originalReadme.replace(/^.*/s, "");

dkarnutsch: oh, wow. I should do more programming :D fixed in b1af899c4983751ef720a5b8d7445c7450afb43a

johnnyomair (dismissed):

thomasdax98 (dismissed):

nsams (dismissed):

johnnyomair (approved):

PR #126, opened 2024-01-25
@comet/create-app: Return non-zero exit code on error

johnnyomair:


create-app/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { Command } from "commander";
+import { Command, program } from "commander";
```
johnnyomair, line 1: Using the global program so that output configuration works in all files/commands. Would need to pass the program variable otherwise

thomasdax98 (dismissed):

nsams:
So `program.error()` does `process.exit()` and aborts the current process without having you to return or throw an error?

I'm not sure If I should like this pattern.

And using this global `program` somewhere in the business code is also not "the yellow form the egg". You could argue, you are doing it in in the outer level function (the one that gets called by the command action) only

nsams:
> I've moved the error handling to the entry point in https://github.com/vivid-planet/comet-starter/commit/fd4640dd4ffcee87458a0e19a514680a81d9b47a. Is this better?

yes (you could even return here, using program.error seems also ok)

> but that seems like exceptions for flow control to me.

yes, not good for valid code paths (like executing the command in a non-comet-cwd)

but perfect for not expected exceptions, like git clone failing

nsams (approved):

dkarnutsch (approved):

johnnyomair:
Merged main to resolve conflicts.

PR #128, opened 2024-01-29
COM-363: Update Starter@next to latest canary release

fraxachun:
- Implement UserPermissions-module
- Use updated SitesConfig-type

(Replacement for https://github.com/vivid-planet/comet-starter/pull/127)


admin/src/config.ts
```diff
@@ -25,4 +27,6 @@ export function createConfig() {
     };
 }
 
+export type SitesConfig = Record<string, SiteConfig>;
```
johnnyomair, line 30: You could add a type cast to `JSON.parse(environmentVariables.SITES_CONFIG)`, which would allow SitesConfigProvider to infer the type

api/src/app.module.ts
```diff
@@ -67,12 +66,20 @@ export class AppModule {
...
-                        return user.domains.includes(requestScope.domain);
-                    },
+                AuthModule.forRoot(config),
+                UserPermissionsModule.forRootAsync({
+                    useFactory: (userService: UserService, accessControlService: AccessControlService) => ({
+                        availablePermissions: ["products"],
+                        availableContentScopes: [
+                            { domain: "main", language: "de" },
+                            { domain: "main", language: "en" },
+                            { domain: "secondary", language: "en" },
```
johnnyomair, line 76:
```suggestion
                            { domain: "secondary", language: "de" },
                            { domain: "secondary", language: "en" },
```

api/src/app.module.ts
```diff
@@ -67,12 +66,20 @@ export class AppModule {
...
+                        ],
+                        userService,
+                        accessControlService,
+                    }),
+                    inject: [UserService, AccessControlService],
+                    imports: [AuthModule.forRoot(config)],
```
johnnyomair, line 82: Doesn't this create two auth modules, and therefore two instances of all services?

api/src/auth/user.service.ts
```diff
@@ -0,0 +1,18 @@
+import { FindUsersArgs, User, UserPermissionsUserServiceInterface, Users } from "@comet/cms-api";
+import { Injectable } from "@nestjs/common";
+
+import { staticUsers } from "./static-users";
+
+@Injectable()
+export class UserService implements UserPermissionsUserServiceInterface {
```
johnnyomair, line 7: Should we add a note here that this needs to be adapted in the application?

create-app/package.json
```diff
@@ -37,7 +37,7 @@
         "rimraf": "^5.0.1"
     },
     "devDependencies": {
-        "@comet/eslint-config": "6.0.0-canary-20231122070819",
+        "@comet/eslint-config": "6.0.0-canary-20240126133306",
```
johnnyomair, line 40: You need to run install in this directory please

package-lock.json
```diff
@@ -17,34 +17,35 @@
             }
         },
         "node_modules/@babel/code-frame": {
-            "version": "7.18.6",
```
johnnyomair, line 20: Why do you have changes in the root package-lock.json, but not in package.json?

api/src/auth/user.service.ts
nsams: This service needs to be named UserLocalService or so and only be used by AuthLocal.

And what default user service for non-local could the starter have?

package-lock.json
johnnyomair: It's not necessary for updating packages, but it doesn't hurt anyway. Would only be a problem if you're accidentally using the wrong NPM version.

api/src/auth/user.service.ts
johnnyomair: > And what default user service for non-local could the starter have?

We could implement a basic user management module in the Starter/Demo. Probably will be deleted in most of our projects, but could be part of the showcase. Of course, out of scope of this PR.

johnnyomair: We'd need a guide about how to make the Starter production-ready then. We should discuss this in tomorrow's meeting.

johnnyomair (approved):

api/src/auth/permission.interface.ts
```diff
@@ -0,0 +1,7 @@
+declare module "@comet/cms-api" {
+    interface Permission {
+        products: string;
+    }
+}
+
+export {};
```
dkarnutsch, line 7: why that?

site/package.json
```diff
@@ -26,11 +26,11 @@
     },
     "dependencies": {
         "@comet/cms-site": "6.0.0-canary-20231122070819",
-        "@opentelemetry/api": "^1.4.0",
-        "@opentelemetry/auto-instrumentations-node": "^0.36.1",
-        "@opentelemetry/exporter-trace-otlp-http": "^0.35.1",
-        "@opentelemetry/sdk-node": "^0.35.1",
-        "draft-js": "^0.11.7",
+        "@next/bundle-analyzer": "^12.3.4",
```
dkarnutsch, line 29: is this related to this PR?

dkarnutsch (commented):
Could we directly target the stable version is it is now released?

thomasdax98 (approved):

PR #154, opened 2024-02-05
@comet/create-app: Automatically fix ESLint errors after commands

max-debug022:
Based on: #159 

After the execution of each command eslint autofix is run if the dependencies are installed. Additionally, create-app now installs the project dependencies per default.


create-app/src/util/lintText.ts
```diff
@@ -0,0 +1,17 @@
+import { ESLint } from "eslint";
+
+export async function lintText(content: string, filePath: string): Promise<string> {
+    const eslint = new ESLint({
+        fix: true,
+        cwd: __dirname,
```
nsams, line 6: __dirname is `.../create-app/src/util`, I don't think that is what you want.

https://www.geeksforgeeks.org/difference-between-process-cwd-and-__dirname-in-node-js/

max-debug022: We need that here because of the behavior of the package. Eslint uses the projects eslint-config. As the config uses an external source (extends from @comet/eslint-config) this leads to an error, when the dependencies are not installed (this is the case for the initial create-app). To avoid this, eslint now uses the eslint-config of itself.

thomasdax98:
@max-debug022 Is there a reason why we don't just execute the lint scripts as a last step? (like https://github.com/vivid-planet/comet-upgrade/pull/9/files) Since I didn't review the script initially, I'm not 100% familiar with how it's structured.

create-app/src/util/lintText.ts
nsams: I understand, please add a comment explaining this

max-debug022:
> @max-debug022 Is there a reason why we don't just execute the lint scripts as a last step? (like https://github.com/vivid-planet/comet-upgrade/pull/9/files) Since I didn't review the script initially, I'm not 100% familiar with how it's structured.

@thomasdax98 That is a good idea - I will implement that.

create-app/src/index.ts
```diff
@@ -22,9 +22,10 @@ void (async () => {
     program
         .argument("<projectName>", "Sets the name of the project.")
         .option("-v, --verbose", "Enables extra console logs for verbose output.")
+        .option("-ni, --no-install", "Disables the automatic installation of dependencies.")
```
thomasdax98, line 25: Please also add the new option to the readme

create-app/src/scripts/create-app/createApp.ts
```diff
@@ -19,6 +23,17 @@ export async function createApp(projectConfiguration: ProjectConfiguration) {
...
+        const spinner = createSpinner("Installing project...").spin();
+        try {
+            execSync("sh ./install.sh");
+            spinner.success();
+            console.log("Installation completed successfully.");
+        } catch (error) {
+            spinner.error({ text: `An error occurred while installing the project: ${error}` });
+        }
+    }
+    runEslintFix();
```
thomasdax98, line 36: I think `createInitialGitCommit();` should be done after `runEslintFix();`. Now the `lang.ts` file is fixed but I have to manually add a second commit

Bildschirmfoto 2024-02-09 um 09 39 35

max-debug022: Yes you are right, but with install.sh, the husky pre-commit hook is also installed which requires the .git folder (which is deleted in the cleanup process) 

I think I might have to split the function up and make the commit afterwards

create-app/src/index.ts
max-debug022: Fixed in: https://github.com/vivid-planet/comet-starter/pull/154/commits/2c3207d4944032ed0b913e7503d7ae11f5c61db8

create-app/src/scripts/create-app/createApp.ts
max-debug022: Fixed in: https://github.com/vivid-planet/comet-starter/pull/154/commits/3c6549a0b1525f78ead6dfc8518c59d09c1c451e

johnnyomair:
@max-debug022 could you please rebase this PR now that #159 has been merged? Thanks!

max-debug022:
> @max-debug022 could you please rebase this PR now that #159 has been merged? Thanks!

@johnnyomair done!

johnnyomair (changes requested):

create-app/src/scripts/remove-showcase/removeShowcase.ts
```diff
@@ -30,26 +30,18 @@ async function removeFileContent() {
...
-        fix: true,
-    });
     for (const content of contentToRemove) {
         if (!existsSync(content.file)) {
             console.log(kleur.bgYellow(`File: ${content.file} does not exist!`));
             console.log(kleur.bgYellow(`Skipping: ${content.file}...`));
             continue;
         }
-        let fileContent = readFileSync(content.file).toString();
+        let fileContent = readFileSync(content.file, "utf-8").toString();
```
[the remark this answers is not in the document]
max-debug022: Yes, you are right - but in this instance, only the files in `contentToRemove` will be read. Setting the BufferEncoding was done to ensure that the file is read and interpreted as a UTF-8 encoded string explicitly.

create-app/src/util/runEslintFix.ts
```diff
@@ -0,0 +1,27 @@
...
+    for (const microservice of microservices) {
+        if (!fs.existsSync(microservice)) {
+            continue;
+        } else if (!hasDependenciesInstalled(microservice)) {
+            console.warn(`Cannot fix ESLint errors in ${microservice} because it has no dependencies installed`);
+            continue;
+        }
+
+        try {
+            if (microservice !== "api") execSync(`npm --prefix ${microservice} run lint:prepare`);
```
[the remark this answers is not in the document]
max-debug022: What would the prepare script do for api?

create-app/src/util/runEslintFix.ts
```diff
@@ -0,0 +1,27 @@
+import { execSync } from "child_process";
+import fs from "fs";
+
+import { hasDependenciesInstalled } from "./hasDependenciesInstalled";
+
+export function runEslintFix() {
+    const microservices = ["admin", "api", "site"];
+    console.log("Fixing ESLint errors");
```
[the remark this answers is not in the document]
max-debug022: I removed the message [here](https://github.com/vivid-planet/comet-starter/pull/154/commits/faf285b6bfa8e6399fb77f377560fc51758af462) since I don't think this message improved the script anyway. For instance, in create-app it will likely not be seen since it is printed before the "whats next"-messages.

create-app/src/scripts/create-app/createApp.ts
```diff
@@ -19,6 +24,18 @@ export async function createApp(projectConfiguration: ProjectConfiguration) {
     cleanupWorkingDirectory(projectConfiguration.verbose);
     replacePlaceholder(projectConfiguration.projectName, projectConfiguration.verbose);
     createInitialGitCommit();
+    if (projectConfiguration.install) {
+        const spinner = createSpinner("Installing project...").spin();
+        try {
+            execSync("sh ./install.sh");
```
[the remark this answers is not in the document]
max-debug022: I tried to fix that issue, but I am not able to. My approach was configuring the child process with a killSignal as recommended in the node docs but this does not seem to work.

@nsams maybe you know another way to implement that?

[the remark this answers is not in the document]
nsams: I thought this is not an issue, as the child process also gets the SIGINT signal sent if you hit Ctrl+C.

Did you test tist?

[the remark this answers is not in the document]
max-debug022: Yes, the activity monitor indicated that the install.sh process **remained active** even after terminating it with Ctrl+C. However, it eventually stopped after 10-15 secs, which is too little time to complete the installation of all dependencies.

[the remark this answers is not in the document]
max-debug022: Yes thats possible. Anyways, after some more testing https://github.com/vivid-planet/comet-starter/pull/154/commits/b553afdf88a66fe57d764e804b26c8d3fcd4134e works.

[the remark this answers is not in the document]
nsams: TLDR: your change is not needed, please revert

killSignal changes the signal sent to the child process when the parent process sends a kill signal because of timeout (if configured) or abort signal (if configured) ([reference](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options)) 

Both are not used in your case and the SIGINT is sent by the terminal where you press Ctrl+C to both processes.

I made a simple test:
```
niko@niko:~/dev/killtest$ cat parent.js
const { execSync } = require("child_process");
execSync("node child.js > child.log 2>&1");
niko@niko:~/dev/killtest$ cat child.js 
const delay = ms => new Promise(resolve => setTimeout(resolve, ms))

process.on('SIGINT', () => {
    console.log("child caught SIGINT, exiting");
    process.exit();
});
process.on('SIGTERM', () => {
    console.log("child caught SIGTERM, exiting");
    process.exit();
});


async function main() {
    console.log('child sleeping for 10 seconds...');
    await delay(10000);
    console.log('child exiting');
}
main();
niko@niko:~/dev/killtest$ node parent.js
^C
niko@niko:~/dev/killtest$ cat child.log 
child sleeping for 10 seconds...
child caught SIGINT, exiting
```

here you see that:
- child process gets SIGINT signal and exits right after pressing Ctrl+C
- it gets SIGINT without having to specify killSignal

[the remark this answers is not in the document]
max-debug022: Done - thank you for clarifying. I appreciate your explanation and code example very much, it helped me to understand the killSignals better. 

Nevertheless, I noticed some strange (most likely not strange, but I don't understand why) behaviors: When adding the killSignal-option, there was only one npm install process. If I am correct, there should be three, as of the install.sh script, which uses the & operator for parallelization. (There are three instances without the option)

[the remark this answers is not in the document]
nsams: The killSignal option has no effect if you don't use timeout option or abort signal option.

You might be right with the & in install.sh, as that creates background processes that _might_ not get the Ctrl+C SIGINT. killSignal doesn't change that though. Is the & worth it anyway?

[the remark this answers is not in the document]
max-debug022: I measured the time for both variants and it does not make a difference. And ctrl+c also works without the killSignal for both variants (It just takes about 15 more seconds)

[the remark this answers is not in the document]
nsams: then I would remove the & to simplify things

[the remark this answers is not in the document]
johnnyomair: > I measured the time for both variants and it does not make a difference

Did you try this on fresh copies? I can't believe that running the npm installs in parallel doesn't make a difference 🤔

create-app/src/util/runEslintFix.ts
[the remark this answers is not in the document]
johnnyomair: You're right, haven't thought about that

create-app/src/scripts/create-app/createApp.ts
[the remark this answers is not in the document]
max-debug022: Yes, I did. I believe this issue is caused by the significantly longer duration of the admin installation compared to the quicker api and site installations. Therefore, when the 'install.sh' script waits for the completion of the admin script, it's almost as if the installations are being done sequentially. On top of that, differences in the download speed add "noise" in the duration, making it such that no noticeable difference is can be seen.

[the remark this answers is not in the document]
johnnyomair: Okay, then remove the "optimization" please 😁

create-app/src/util/hasDependenciesInstalled.ts
```diff
@@ -0,0 +1,10 @@
+import fs from "fs";
+
+export function hasDependenciesInstalled(microservice: string): boolean {
+    const files = fs.readdirSync(`${process.cwd()}/${microservice}`);
+    if (!files.includes("package.json")) {
+        return false;
+    }
+    const packageJson = fs.readFileSync(`${process.cwd()}/${microservice}/package.json`, "utf8");
+    return (packageJson.includes("dependencies") || packageJson.includes("devDependencies")) && fs.existsSync("node_modules");
```
thomasdax98, line 9: In line 8 you use the absolute path
```
 const packageJson = fs.readFileSync(`${process.cwd()}/${microservice}/package.json`, "utf8");
```
here you use the relative path
```
fs.existsSync("node_modules");
```
Are you sure this works as expected?

thomasdax98: otherwise lgtm

thomasdax98:
And please resolve the conflicts

create-app/src/util/hasDependenciesInstalled.ts
nsams: that is practically the same, although `${process.cwd()}/` could be omitted

create-app/src/scripts/create-app/createApp.ts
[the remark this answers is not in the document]
max-debug022: Changed in: https://github.com/vivid-planet/comet-starter/pull/154/commits/7a52dd70c0909e2c16a9c740349c4614e0bd72d0

create-app/src/util/hasDependenciesInstalled.ts
max-debug022: Yes, it works. I prefer the absolute path when changing non-project files because:
- Readability/Clarity
- Error Prevention
- Avoidance of Ambiguity

admin/package.json
```diff
@@ -8,7 +8,8 @@
         "serve": "cd server && npm run serve",
         "gql:types": "graphql-codegen",
         "gql:watch": "graphql-codegen --watch",
-        "lint": "npm run intl:compile && run-p gql:types generate-block-types && run-p lint:prettier lint:eslint lint:tsc",
+        "lint": "npm run lint:prepare && run-p lint:eslint lint:tsc",
+        "lint:prepare": "npm run intl:compile && run-p gql:types generate-block-types",
```
dkarnutsch, line 12: if you don't want to this manually, you can use https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-scripts

thomasdax98 (approved):

create-app/src/util/hasDependenciesInstalled.ts
max-debug022: Oh that's what you meant - Sorry I misunderstood that. No in that case you are right, I intended to check for the microservices node_modules (root also has some, so it works anyway)

admin/package.json
max-debug022: Fixed in: https://github.com/vivid-planet/comet-starter/pull/154/commits/b2f5f2f0f0aef7d5aed4fd4e88c9116571516e77

create-app/src/util/hasDependenciesInstalled.ts
max-debug022: Solved in: https://github.com/vivid-planet/comet-starter/pull/154/commits/d4b5176b4e155f41d1d49ad510d2b65c88052088

admin/package.json
johnnyomair: This doesn't really change much, since you still need to call `prelint` manually before running ESLint 🤔

johnnyomair:
@max-debug022 please fix the failing pipeline

admin/package.json
max-debug022: But now it's the recommended way - isn't that preferable?

johnnyomair: Of course. I'd still prefer if we could remove the manual run and the check for api in `runEslintFix` somehow

johnnyomair (approved):

thomasdax98 (approved):

dkarnutsch (approved):

nsams (approved):

PR #95, opened 2024-01-03
Enhance linting

dkarnutsch:
Following problems are addressed here:
- Lint-Staged should match actual lint config
- Linting project root did not take into account subfolders (i.e. .gitlab ci) due to misconfigured glob pattern
- Yaml files did not get linted (e.g. helm files) due to missing extension
- Files outside of folder "src" from microservices (api/admin/site) did not get linted as eslint did only get executed on "src" folder (e.g. next-config or readme)
- Harmonize lint calls with run-p

However, this has bad IDE support as we have a prettierignore with (api/admin/site/create-app). This can be resolved by either copying the .gitignore into .prettierignore and removing the ignore patterns or by updating to prettier 3 (which ignores all files from .gitignore by default, https://prettier.io/blog/2023/07/05/3.0.0.html#cli)

PR #169, opened 2024-02-23
Allow Prettier in IDE for microservices

fraxachun:
https://github.com/vivid-planet/comet-starter/pull/95#issue-2064259483

Fixes the "bad IDE support" by allowing allowing prettier in IDE for api/admin/site/create-app but executing the cli-check only for root-files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
         "dev": "dotenv -c secrets -- dev-pm start",
         "copy-schema-files": "node copy-schema-files.js",
         "lint": "run-p lint:*",
-        "lint:root": "npx prettier --check './*.{js,json,md,yml,yaml}'",
+        "lint:root": "npx prettier --check './!(admin|api|site|create-app)/**/*.{js,json,md,yml,yaml}'",
         "lint:api": "npm --prefix api run lint",
         "lint:admin": "npm --prefix admin run lint",
         "lint:site": "npm --prefix site run lint",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
         "dev": "dotenv -c secrets -- dev-pm start",
         "copy-schema-files": "node copy-schema-files.js",
         "lint": "run-p lint:*",
-        "lint:root": "npx prettier --check './**/*.{js,json,md,yml,yaml}'",
```
dkarnutsch, line 8: This does not work for values.yaml in chart/.

We would need to exclude admin/api/site/create-app manually.

dkarnutsch (changes requested):

package.json
```diff
@@ -5,7 +5,7 @@
...
+        "lint:root": "npx prettier --check './!(admin|api|site|create-app)/**/*.{js,json,md,yml,yaml}'",
```
nsams, line 8: now that this doesn't lint root alone anymore, maybe rename to:
```suggestion
        "lint:prettier": "npx prettier --check './!(admin|api|site|create-app)/**/*.{js,json,md,yml,yaml}'",
```

nsams: or, wait: des the `!` mean those will be excluded?

nsams: ok, so why this change? does it still fix "bad IDE support"?

nsams (approved):

johnnyomair (approved):
Finally, thanks!

johnnyomair:
@dkarnutsch please review again, thanks!

dkarnutsch (approved):

thomasdax98 (approved):

PR #181, opened 2024-03-04
Create rewrites to avoid redirect loop

magdalenaxm:
Creates rewrites for redirect loop, as case sensitivity in Next.js is inconsistent (see also https://github.com/vercel/next.js/issues/21498#issuecomment-848315717).

Background:

Sometimes you want a redirect from an uppercase to a lowercase URL (e.g. /Example to /example) for SEO purposes. This can be created in our admin without problems. The casing is saved correctly in the DB and transferred to Next.

The problem is that this creates a redirect loop in the site. The reason is that the case sensitivity of Next.js is inconsistent. The delivery of pages is case sensitive. The redirects are case insensitive:

Without redirect:

http://localhost:3000/example -> page is delivered
http://localhost:3000/Example -> 404

With redirect:

http://localhost:3000/example -> Redirect to /example -> Loop
http://localhost:3000/Example -> Redirect to /example -> Loop

Also described here: https://github.com/vercel/next.js/issues/21498

### DIFF
--- a/site/next.config.js
+++ b/site/next.config.js
@@ -42,7 +42,7 @@ const nextConfig = {
     },
     redirects: async () => {
         if (process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true") return [];
-        var { redirects } = await require("./preBuild/build/preBuild/src/createRedirects").createRedirects();
+        var redirects = await require("./preBuild/build/preBuild/src/createRedirects").createRedirects();
         return redirects;
     },
     images: {
--- a/site/next.config.js
+++ b/site/next.config.js
@@ -37,7 +37,7 @@ const nextConfig = {
     basePath: process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true" ? "/site" : "",
     rewrites: async () => {
         if (process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true") return [];
-        var { rewrites } = await require("./preBuild/build/preBuild/src/createRedirects").createRedirects();
+        var rewrites = await require("./preBuild/build/preBuild/src/createRewrites").createRewrites();
         return rewrites;
     },
     redirects: async () => {
--- a/site/next.config.js
+++ b/site/next.config.js
@@ -35,9 +35,14 @@ const nextConfig = {
     pageExtensions: ["page.ts", "page.tsx"],
     cleanDistDir: process.env.NODE_ENV !== "production", // sitemap and robots.txt are pre-existing
     basePath: process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true" ? "/site" : "",
+    rewrites: async () => {
+        if (process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true") return [];
+        var { rewrites } = await require("./preBuild/build/preBuild/src/createRedirects").createRedirects();
+        return rewrites;
+    },
     redirects: async () => {
         if (process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true") return [];
-        var redirects = await require("./preBuild/build/preBuild/src/createRedirects").createRedirects();
+        var { redirects } = await require("./preBuild/build/preBuild/src/createRedirects").createRedirects();
         return redirects;
     },
     images: {
--- a/site/preBuild/src/createRedirects.ts
+++ b/site/preBuild/src/createRedirects.ts
@@ -89,7 +89,7 @@ const createApiRedirects = async (): Promise<Redirect[]> => {
             continue;
         }
 
-        if (source && destination && source.toLowerCase() !== destination.toLowerCase()) {
+        if (source && destination) {
             redirects.push({ source, destination, permanent: true });
         }
     }
--- a/site/preBuild/src/createRedirects.ts
+++ b/site/preBuild/src/createRedirects.ts
@@ -24,7 +24,7 @@ const redirectsQuery = gql`
     }
 `;
 
-async function* getRedirects() {
+export async function* getRedirects() {
     let offset = 0;
     const limit = 100;
 
@@ -87,21 +87,13 @@ const createApiRedirects = async (): Promise<{ redirects: Redirect[]; rewrites: 
             }
         }
 
-        if (source === destination) {
+        if (source?.toLowerCase() === destination?.toLowerCase()) {
             console.warn(`Skipping redirect loop ${source} -> ${destination}`);
             continue;
         }
 
-        if (source && destination) {
-            if (source.toLowerCase() === destination.toLowerCase()) {
-                const newSource = source
-                    .split("")
-                    .map((char) => (char.toLowerCase() === char.toUpperCase() ? char : `(${char.toLowerCase()}|${char.toUpperCase()})`))
-                    .join("");
-                rewrites.push({ source: newSource, destination });
-            } else {
-                redirects.push({ source, destination, permanent: true });
-            }
+        if (source && destination && source.toLowerCase() !== destination.toLowerCase()) {
+            redirects.push({ source, destination, permanent: true });
         }
     }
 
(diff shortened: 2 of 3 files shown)


thomasdax98 (dismissed):

thomasdax98:
Demo PR: https://github.com/vivid-planet/comet/pull/1787

site/next.config.js
```diff
@@ -35,9 +35,14 @@ const nextConfig = {
     pageExtensions: ["page.ts", "page.tsx"],
     cleanDistDir: process.env.NODE_ENV !== "production", // sitemap and robots.txt are pre-existing
     basePath: process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true" ? "/site" : "",
+    rewrites: async () => {
+        if (process.env.NEXT_PUBLIC_SITE_IS_PREVIEW === "true") return [];
+        var { rewrites } = await require("./preBuild/build/preBuild/src/createRedirects").createRedirects();
```
johnnyomair, line 40: Since you're fetching the API redirects twice anyway, I see no benefit in returning the rewrites in createRedirects (expect for less LOC). I'd propose splitting createRedirects into three functions:

1. `fetchRedirects`: Fetches the redirects using the GraphQL API
2. `createRedirects`: Creates the redirects, but skips those which are equal except for casing. Something like `fetchRedirects().filter(({ source, destination }) => source.toLowerCase() !== destination.toLowerCase())`
3. `createRewrites`: Creates the rewrites for case-sensitive redirects. Something like `fetchRedirects().filter(({ source, destination }) => source.toLowerCase() === destination.toLowerCase()).map(() => /* Rewrite source */)`

This would separate the concerns more clearer. What do you think?

site/preBuild/src/createRedirects.ts
```diff
@@ -90,11 +93,19 @@ const createApiRedirects = async (): Promise<Redirect[]> => {
         }
 
         if (source && destination) {
-            redirects.push({ source, destination, permanent: true });
+            if (source.toLowerCase() === destination.toLowerCase()) {
+                const newSource = source
+                    .split("")
+                    .map((char) => (char.toLowerCase() === char.toUpperCase() ? char : `(${char.toLowerCase()}|${char.toUpperCase()})`))
+                    .join("");
```
johnnyomair, line 100: Is this necessary? Shouldn't a rewrite from "/Example" to "/example" be enough? Or do we want to support "/ExAmplE"?

thomasdax98: Yes, you are right. Actually we can simply do `/Example` to `/example` so it behaves like the redirect would behave.

site/next.config.js
magdalenaxm: I think you're right. I've updated it here: ee224492642e20b422aa7104d316285b34dfbc59

site/preBuild/src/createRedirects.ts
magdalenaxm: I've updated here: ee224492642e20b422aa7104d316285b34dfbc59

site/preBuild/src/createRedirects.ts
```diff
@@ -53,14 +55,15 @@ const createInternalRedirects = async (): Promise<Redirect[]> => {
         },
     ];
 };
-const createApiRedirects = async (): Promise<Redirect[]> => {
+const createApiRedirects = async (): Promise<{ redirects: Redirect[]; rewrites: Rewrite[] }> => {
```
johnnyomair, line 58: This should only return redirects now

site/preBuild/src/createRedirects.ts
```diff
@@ -84,17 +87,17 @@ const createApiRedirects = async (): Promise<Redirect[]> => {
...
         }
 
-        if (source === destination) {
+        if (source?.toLowerCase() === destination?.toLowerCase()) {
             console.warn(`Skipping redirect loop ${source} -> ${destination}`);
             continue;
         }
 
-        if (source && destination) {
+        if (source && destination && source.toLowerCase() !== destination.toLowerCase()) {
```
johnnyomair, line 95: You already have to check on casing in line 90

site/preBuild/src/createRewrites.ts
```diff
@@ -0,0 +1,49 @@
...
+            switch (target.block.type) {
+                case "internal":
+                    destination = (target.block.props as InternalLinkBlockData).targetPage?.path;
+                    break;
+
+                case "external":
+                    destination = (target.block.props as ExternalLinkBlockData).targetUrl;
+                    break;
+            }
+        }
```
johnnyomair, line 35: Could we move this into getRedirects?

site/preBuild/src/createRedirects.ts
magdalenaxm: fe80ca8648db34206cf966e18e07c451c4ff6e64

site/preBuild/src/createRewrites.ts
magdalenaxm: should this be a helper method which is used by both, createRewrites and createRedirects?

johnnyomair: I'd just make it part of fetching the redirects (in getRedirects), since it's the same for both use cases, right?

site/preBuild/src/createRedirects.ts
magdalenaxm: 156ce7dea8f28cd0b333af148fd5d4414fc98220

site/preBuild/src/createRewrites.ts
magdalenaxm: 9048d16bbc04cbf0585c224dafdc3a33a7d6fae9

site/preBuild/src/createRewrites.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    for await (const redirect of getRedirects()) {
+        const { source, destination } = redirect;
+
+        if (source && destination && source.toLowerCase() === destination.toLowerCase()) {
+            const firstChar = source.charAt(1);
+
+            const pattern = new RegExp(`(${firstChar.toLowerCase()}|${firstChar.toUpperCase()})`, "g");
+
+            rewrites.push({ source: source.replace(pattern, "($1)"), destination });
```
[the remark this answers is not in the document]
magdalenaxm: 6c97a242ca2c4f220b94a411ddabf71c2c936b64

site/preBuild/src/createRewrites.ts
```diff
@@ -0,0 +1,25 @@
...
+        return { redirects: [], rewrites: [] };
+    }
+
+    const rewrites: Rewrite[] = [];
+
+    for await (const redirect of getRedirects()) {
+        const { source, destination } = redirect;
+
+        if (source && destination && source.toLowerCase() === destination.toLowerCase()) {
+            rewrites.push({ source, destination });
```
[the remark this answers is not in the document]
magdalenaxm: 8dad1b7243f61ba90706857961b06bcfdb51091e

site/preBuild/src/createRewrites.ts
```diff
@@ -0,0 +1,27 @@
...
+        return { redirects: [], rewrites: [] };
+    }
+
+    const rewrites: Rewrite[] = [];
+
+    for await (const redirect of getRedirects()) {
+        const { source, destination } = redirect;
+
+        /* A rewrite is created for each redirect where source and destination differ only by casing (otherwise this causes a redirection loop).
+        E.g. a rewrite is created for the redirect /Example -> /example*/
```
johnnyomair, line 18:
```suggestion
        // A rewrite is created for each redirect where the source and destination differ only by casing (otherwise, this causes a redirection loop).
        // For instance, a rewrite is created for the redirect /Example -> /example.
```

johnnyomair (approved):

thomasdax98 (approved):

nsams (approved):

PR #194, opened 2024-03-11
@comet/create-app: Support Comet v6

max-debug022:
This PR adds following changes:
- Support for @comet/create-app in Comet v6

### DIFF
--- a/create-app/src/scripts/remove-showcase/removeShowcase.ts
+++ b/create-app/src/scripts/remove-showcase/removeShowcase.ts
@@ -11,7 +11,7 @@ async function removeFileContent() {
     }> = [
         {
             file: `./admin/src/common/masterMenuData.tsx`,
-            replacements: [/.{217}\.products.*?,\s{5}},\n/gs, /.*Products.*/gm],
+            replacements: [/\s*\/\/ This.*\.products.*?,\s{5}},/gs, /.*Products.*/gm],
         },
         {
             file: `api/src/auth/permission.interface.ts`,
--- a/create-app/src/scripts/remove-showcase/removeShowcase.ts
+++ b/create-app/src/scripts/remove-showcase/removeShowcase.ts
@@ -10,16 +10,16 @@ async function removeFileContent() {
         replacements: Array<string | RegExp>;
     }> = [
         {
-            file: `./admin/src/Routes.tsx`,
-            replacements: [/\s*<.*?products.*?>/g],
+            file: `./admin/src/common/masterMenuData.tsx`,
+            replacements: [/.{217}\.products.*?,\s{5}},\n/gs, /.*Products.*/gm],
         },
         {
-            file: `./admin/src/common/MasterMenu.tsx`,
-            replacements: [/\s*<MenuCollapsibleItem.*products`}\s*\/>\s*<\/MenuCollapsibleItem>/gs],
+            file: `api/src/auth/permission.interface.ts`,
+            replacements: [/\n.*products.*?}\n/gs],
         },
         {
             file: `./api/src/app.module.ts`,
-            replacements: [/\s*ProductsModule,/gs],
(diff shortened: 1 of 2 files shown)


create-app/src/scripts/remove-showcase/removeShowcase.ts
```diff
@@ -10,16 +10,16 @@ async function removeFileContent() {
         replacements: Array<string | RegExp>;
     }> = [
         {
-            file: `./admin/src/Routes.tsx`,
-            replacements: [/\s*<.*?products.*?>/g],
+            file: `./admin/src/common/masterMenuData.tsx`,
+            replacements: [/.{217}\.products.*?,\s{5}},\n/gs, /.*Products.*/gm],
```
johnnyomair, line 14: Isn't there a better regex for this? Checking for exactly 217 characters could break rather easily.

johnnyomair (commented):
This PR should be two PRs. Both changes are in the `@comet/create-app` scope, but unrelated

create-app/src/scripts/remove-showcase/removeShowcase.ts
max-debug022: I agree - But I couldn't come up with a better regex for this particular code.

nsams:
create-app should not break when starter changes. We could add a pipeline that verifies the replacements can be applied without errors

create-app/src/scripts/remove-showcase/removeShowcase.ts
nsams: anything is better than character counting

:two: is ok

max-debug022: Fixed with: https://github.com/vivid-planet/comet-starter/pull/194/commits/faf87bf2821f586ee525fbaf04d9b8cc97c7b098

nsams: I thought more about something like that:
```
// BEGIN product-showcase This line is needed for @comet/create-app (Do not remove this line)
....code that will be removed
// END product-showcase This line is needed for @comet/create-app (Do not remove this line)
```
you can apply that pattern on all files

max-debug022: How would you use that pattern with files like `fixtures.module.ts`, where content within a single line is deleted?

nsams: Either

- leave that like it is now (and working perfectly) and use the comment technique only for not-so-easy areas, maybe for now only masterMenuData
- or add the comment everywhere and split into multiple lines:
```
    imports: [
        // BEGIN product-showcase This line is needed for @comet/create-app (Do not remove this line)
        MikroOrmModule.forFeature([Product]),
        // END product-showcase This line is needed for @comet/create-app (Do not remove this line)
        ConfigModule,
        ConsoleModule
    ],
```

that would add a lot of comments though, maybe :one: would be better

max-debug022: I think 1 might be better because if we add comments everywhere, removeShowcase could be simplified (the script would only look for this comment).
Pro:
- Not prone to regex-errors 
- PRs like this one would not be necessary in the future hence we only update the comments
Con:
- More comments

nsams: fine with me too

johnnyomair: I like the Begin/End approach better, because it's more explicit. However, we should first discuss if we even need the showcase in the Starter

johnnyomair:
I believe we can close this in favor of #206.

max-debug022:
Closed in favor of https://github.com/vivid-planet/comet-starter/pull/206.

PR #198, opened 2024-03-12
Add document type loader as in comet demo

nsams:
adapt changes from  https://github.com/vivid-planet/comet/pull/1338 and https://github.com/vivid-planet/comet/pull/1815


site/src/documents/pages/index.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
```
johnnyomair, line 1: This should be in `documents/`, not `documents/pages/`

johnnyomair (changes requested):

site/src/documents/pages/index.tsx
johnnyomair: Sorry, I only meant the types for the document types, not the page document type itself 😅 That's fine in documents/pages/

PR #200, opened 2024-03-14
Use API Generator for products

johnnyomair:


api/schema.gql
```diff
@@ -163,12 +163,10 @@ type PaginatedUserList {
 
 type Product {
   id: ID!
-  creatorId: String!
   name: String!
   description: String!
   createdAt: DateTime!
   updatedAt: DateTime
-  sales: Int!
```
johnnyomair, line 171: This was a field resolver that always returned 0. I see no use in it other than demonstrating how to create a field resolver. Should I still add it?

nsams (approved):

api/schema.gql
thomasdax98: > Should I still add it?

I'd say no

thomasdax98 (approved):

dkarnutsch (approved):

PR #206, opened 2024-03-20
Remove products showcase

johnnyomair:
We decided that the products showcase isn't relevant anymore since we're generating API/Admin anyway. The generated code should be considered as best practice instead.


api/src/products/products.console.ts
```diff
@@ -1,22 +0,0 @@
...
-    private readonly logger = new Logger(ProductsConsole.name);
-
-    constructor(
-        private readonly orm: MikroORM, // MikroORM is injected so we can use the request context
-    ) {}
-
-    @Command({
-        command: "demo-command",
-        description: "Demo-Command for cronjob-deployment",
-    })
```
johnnyomair, line 17: Should we add another demo command instead of this?

api/src/db/fixtures/fixtures.console.ts
```diff
@@ -41,7 +37,7 @@ export class FixturesConsole {
         await migrator.up();
 
         const multiBar = new MultiBar(this.barOptions, Presets.shades_classic);
-        await Promise.all([generateProducts({ repository: this.productsRepository, bar: multiBar.create(total, 0), total })]);
+        // Add your fixtures here
```
johnnyomair, line 40: We should add page tree fixtures here.

api/src/products/products.console.ts
max-debug022: I would prefer detailed documentation over demo code as you should delete the demo code anyways when deploying to a productive environment.

I think Comet Starter should be a ready-to-go setup as the demo on the other hand may contain all demo examples.

nsams (dismissed):

thomasdax98 (approved):

johnnyomair:
@max-debug022 @fraxachun please review.

max-debug022 (approved):

PR #215, opened 2024-04-09
Disable GraphQL field suggestions in production

jomunker:
Add `formatError` property to `GraphQLModule` options to hide `GraphQL` field suggestions for non-dev-environments.

### DIFF
--- a/api/src/app.module.ts
+++ b/api/src/app.module.ts
@@ -15,7 +15,7 @@ import {
     RedirectsModule,
     UserPermissionsModule,
 } from "@comet/cms-api";
-import { ApolloDriver } from "@nestjs/apollo";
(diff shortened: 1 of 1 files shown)


thomasdax98 (approved):

johnnyomair:
> When a CometValidationException is thrown, is this also covered here?

I don't think so, but should it?

> Why not integrating this code directly in our [ExceptionInterceptor](https://github.com/vivid-planet/comet/blob/main/packages/api/cms-api/src/common/errors/exception.interceptor.ts)?

Good idea! @jomunker could you please give it a try?

johnnyomair (approved):

dkarnutsch (approved):

nsams (approved):

PR #216, opened 2024-04-15
Access Admin config via dependency injection

johnnyomair:
The config shouldn't be created anew everytime it is needed in a component. Instead, it should be accessed via dependency injection. To achieve that, we add a `ConfigProvider` and `useConfig` hook.

---

COM-638

### DIFF
--- a/admin/src/App.tsx
+++ b/admin/src/App.tsx
@@ -54,45 +54,45 @@ const apiClient = createHttpClient(config.apiUrl);
 
 export function App() {
     return (
-        <ApolloProvider client={apolloClient}>
-            <CurrentUserProvider>
-                <BuildInformationProvider value={{ date: config.buildDate, number: config.buildNumber, commitHash: config.commitSha }}>
-                    <SitesConfigProvider
-                        value={{
-                            configs: config.sitesConfig,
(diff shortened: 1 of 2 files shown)


johnnyomair:
We discussed adding a warning to `createConfig` if it's called multiple times, but that doesn't work when using Vite and Hot Reloading, since it caused a lot of false positives when applying changes. I therefore decided to not add it.

admin/src/App.tsx
```diff
@@ -92,30 +92,32 @@ export function App() {
...
-                                                                                <MasterMenuRoutes menu={masterMenuData} />
-                                                                            </MasterLayout>
-                                                                        )}
-                                                                    />
-                                                                </Switch>
-                                                            )}
-                                                        </ContentScopeProvider>
-                                                        <ErrorDialogHandler />
-                                                    </RouterBrowserRouter>
+                                                    <ConfigProvider config={config}>
```
dkarnutsch, line 95: Could the provider be more on top? I don't think it needs to rerender.

thomasdax98 (dismissed):

nsams (approved):

dkarnutsch (approved):

PR #217, opened 2024-04-15
Remove unnecessary React imports in Admin

johnnyomair:
The React import doesn't have to be in the JSX scope anymore since we're now using the new transform.


dkarnutsch (approved):
I like that we also use functions for React components as our best practice even though we don't enforce it

mennoxx (approved):

thomasdax98 (approved):

nsams (approved):

PR #222, opened 2024-04-16
Create block index views in `FixturesConsole`

thomasdax98:
Corresponds to https://github.com/vivid-planet/comet/pull/1966

The block index views must be (re)created 

- after executing database migrations
- after executing the fixtures (because they drop the whole database and recreate it)

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -22,7 +22,7 @@
         "mikro-orm:migration:generate": "mikro-orm migration:create",
         "start": "npm run prebuild && dotenv -c secrets -- nest start",
         "start:debug": "npm run prebuild && dotenv -c secrets -- nest start --debug --watch --preserveWatchOutput",
-        "start:dev": "npm run prebuild && npm run db:migrate && npm run console createBlockIndexViews && dotenv -c secrets -- nest start --watch --preserveWatchOutput",
+        "start:dev": "npm run prebuild && npm run db:migrate && dotenv -c secrets -- nest start --watch --preserveWatchOutput",
         "start:prod": "node dist/main",
         "test": "jest --passWithNoTests",
         "test:ci": "npm run test -- --ci --reporters=default --reporters=jest-junit",
--- a/api/package.json
+++ b/api/package.json
@@ -22,7 +22,7 @@
         "mikro-orm:migration:generate": "mikro-orm migration:create",
         "start": "npm run prebuild && dotenv -c secrets -- nest start",
         "start:debug": "npm run prebuild && dotenv -c secrets -- nest start --debug --watch --preserveWatchOutput",
-        "start:dev": "npm run prebuild && npm run db:migrate && dotenv -c secrets -- nest start --watch --preserveWatchOutput",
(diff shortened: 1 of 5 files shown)


api/package.json
```diff
@@ -10,7 +10,9 @@
         "db:migrate": "npm run console migrate --",
         "db:migrate:prod": "npm run console:prod migrate --",
         "fixtures": "npm run console fixtures --",
+        "postfixtures": "npm run console createBlockIndexViews",
         "fixtures:prod": "npm run console:prod fixtures --",
+        "postfixtures:prod": "npm run console createBlockIndexViews",
```
thomasdax98, line 15: I'm not sure when the prod fixtures are executed and if we need that here?

nsams:
please explain the problem you want to solve and where it happens

api/package.json
johnnyomair: Maybe they are executed in non-prod environments in some projects?

thomasdax98:
TODO: Call create view in fixtures console

api/src/db/migrate.console.ts
```diff
@@ -70,6 +71,7 @@ export class MigrateConsole {
             this.logger.log(`Executed migrations. Trying to commit...`);
             await em.commit();
             this.logger.log("Migrations successfully committed");
+            await this.dependenciesService.createViews();
```
nsams, line 74: in production, the createViews is done in a separate Job

- if you change that here, that job must be removed
- if i remember correctly, the reason for the job is that it doesn't block api startup (and can run in background) -  migrations must block

api/src/db/fixtures/fixtures.console.ts
```diff
@@ -40,6 +41,8 @@ export class FixturesConsole {
         // Add your fixtures here
         multiBar.stop();
 
+        await this.dependenciesService.createViews();
```
nsams, line 44: :+1:

api/src/db/migrate.console.ts
```diff
@@ -70,6 +71,10 @@ export class MigrateConsole {
             this.logger.log(`Executed migrations. Trying to commit...`);
             await em.commit();
             this.logger.log("Migrations successfully committed");
+
+            if (process.env.NODE_ENV === "development") {
+                await this.dependenciesService.createViews();
```
johnnyomair, line 76: > after executing database migrations

Why? Because of possible new entities with root blocks?

thomasdax98: Yes

johnnyomair (dismissed):

api/src/db/migrate.console.ts
nsams: It is not very obvious that migrate command does under a NODE_ENV condition a createViews. I'd keep that like it was before clearly separated and handled in package.json.

dkarnutsch (approved):

johnnyomair (approved):

dkarnutsch (approved):

api/src/db/migrate.console.ts
thomasdax98: Done in bab79e59fae4c0f83b472ad16ed86b3b20da4fc0

nsams (approved):

PR #223, opened 2024-04-17
Add pageTreeNode field resolver to documents

thomasdax98:
That's necessary for the `createDependencyMethods` helper to work correctly

### DIFF
--- a/api/src/documents/links/links.resolver.ts
+++ b/api/src/documents/links/links.resolver.ts
@@ -1,8 +1,16 @@
-import { AffectedEntity, PageTreeNodeVisibility, PageTreeService, RequiredPermission, validateNotModified } from "@comet/cms-api";
+import {
+    AffectedEntity,
+    PageTreeNodeInterface,
+    PageTreeNodeVisibility,
+    PageTreeService,
+    RequiredPermission,
+    validateNotModified,
+} from "@comet/cms-api";
 import { InjectRepository } from "@mikro-orm/nestjs";
 import { EntityRepository } from "@mikro-orm/postgresql";
 import { UnauthorizedException } from "@nestjs/common";
-import { Args, ID, Mutation, Query, Resolver } from "@nestjs/graphql";
+import { Args, ID, Mutation, Parent, Query, ResolveField, Resolver } from "@nestjs/graphql";
+import { PageTreeNode } from "@src/page-tree/entities/page-tree-node.entity";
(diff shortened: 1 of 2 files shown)


api/src/documents/links/links.resolver.ts
```diff
@@ -17,6 +25,11 @@ export class LinksResolver {
         return this.repository.findOne(linkId);
     }
 
+    @ResolveField(() => PageTreeNode, { nullable: true })
+    async pageTreeNode(@Parent() link: Link): Promise<PageTreeNodeInterface | null> {
+        return this.pageTreeService.createReadApi({ visibility: "all" }).getFirstNodeByAttachedPageId(link.id);
```
thomasdax98, line 30: I argue if I query the pageTreeNode via the document, I always want to get it, no matter the visibility.

johnnyomair (approved):

api/src/documents/links/links.resolver.ts
nsams: The question is not if you need it, but rather if it is potential security issue or not

johnnyomair: Wouldn't it be possible to inject the `PageTreeReadApiService` instead that handles the visibility correctly?

thomasdax98: @nsams I don't think the `PageTreeNode` is the sensitive part of a document. The sensitive part is the content (if it's not published). Am I mistaken?

In this case, the `PageTreeNode` is accessed via the document. If I'm allowed to access the document, I'm allowed to access the `PageTreeNode`. 

If I'm not allowed to access the document because of it's visibility, the query shouldn't be possible at all. But if we want to check that, we must check it in the query, not in this field resolver.

nsams: > I don't think the PageTreeNode is the sensitive part of a document. The sensitive part is the content (if it's not published). Am I mistaken?

You are right, the content is more sensitive than the PageTreeNode. But that does _not_ mean we don't have to protect PageTreeNode!

> In this case, the PageTreeNode is accessed via the document. If I'm allowed to access the document, I'm allowed to access the PageTreeNode.

:+1:

Now we have thought about security and know that this is no potential issue :)

nsams (approved):

dkarnutsch (approved):

PR #237, opened 2024-05-13
Add `HiddenIfInvalidLink` helper

kerstin97:
Hides a link and its content if the link target isn't valid.

---

COM-454
Block-Preview:
image
Web-Preview:
image
Site:
image

### DIFF
--- a/site/src/common/blocks/RichTextBlock.tsx
+++ b/site/src/common/blocks/RichTextBlock.tsx
@@ -4,6 +4,7 @@ import * as React from "react";
 import redraft, { Renderers } from "redraft";
 import styled from "styled-components";
 
+import { HiddenIfInvalidLink } from "../helpers/HiddenIfInvalidLink";
 import { LinkBlock } from "./LinkBlock";
 
 /**
@@ -93,9 +94,11 @@ const defaultRenderers: Renderers = {
         // key is the entity key value from raw
         LINK: (children, data: LinkBlockData, { key }) => {
             return (
-                <LinkBlock key={key} data={data}>
-                    <a>{children}</a>
-                </LinkBlock>
+                <HiddenIfInvalidLink link={data}>
+                    <LinkBlock key={key} data={data}>
+                        <a>{children}</a>
+                    </LinkBlock>
+                </HiddenIfInvalidLink>
             );
         },
     },
--- a/site/src/common/blocks/RichTextBlock.tsx
+++ b/site/src/common/blocks/RichTextBlock.tsx
@@ -4,7 +4,6 @@ import * as React from "react";
 import redraft, { Renderers } from "redraft";
 import styled from "styled-components";
 
-import { HiddenIfInvalidLink } from "../helpers/HiddenIfInvalidLink";
 import { LinkBlock } from "./LinkBlock";
 
 /**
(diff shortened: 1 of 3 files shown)


site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -93,9 +94,11 @@ const defaultRenderers: Renderers = {
...
         LINK: (children, data: LinkBlockData, { key }) => {
             return (
-                <LinkBlock key={key} data={data}>
-                    <a>{children}</a>
-                </LinkBlock>
+                <HiddenIfInvalidLink link={data}>
+                    <LinkBlock key={key} data={data}>
+                        <a>{children}</a>
+                    </LinkBlock>
+                </HiddenIfInvalidLink>
```
johnnyomair, line 101: I don't think using it in RTE is the best place to demonstrate its usefulness: The link's text inside a copy text shouldn't disappear if no valid link is set. This would lead to gaps in the text.

How about demonstrating it in [PageLink](https://github.com/vivid-planet/comet-starter/blob/main/site/src/layout/header/PageLink.tsx#L26) instead?

site/src/common/helpers/HiddenIfInvalidLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { ExternalLinkBlockData, InternalLinkBlockData, LinkBlockData } from "@src/blocks.generated";
+import * as React from "react";
+
+interface HiddenIfInvalidLinkProps {
+    link: LinkBlockData;
+    children: React.ReactElement;
+}
+
+export function HiddenIfInvalidLink({ link: { block }, children }: HiddenIfInvalidLinkProps): React.ReactElement | null {
+    const { hasBridge: isInPreview } = useIFrameBridge();
```
johnnyomair, line 11: Please use `usePreview` and `previewType` instead, which is more explicit IMO

site/src/common/helpers/HiddenIfInvalidLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    if (!isValidLink(block)) {
+        return null;
+    }
+
+    return children;
+}
+
+const isValidLink = (linkBlock: LinkBlockData["block"]) => {
```
johnnyomair, line 24: I'd prefer if we'd work with LinkBlockData here. Maybe this function will be exported in some projects to be used for other scenarios. Then, always having to write `link.block` is cumbersome.

site/src/common/helpers/HiddenIfInvalidLink.tsx
```diff
@@ -0,0 +1,30 @@
+import { usePreview } from "@comet/cms-site";
+import { ExternalLinkBlockData, InternalLinkBlockData, LinkBlockData } from "@src/blocks.generated";
+import * as React from "react";
+
+interface HiddenIfInvalidLinkProps {
+    link: LinkBlockData;
+    children: React.ReactElement;
+}
+
+export function HiddenIfInvalidLink({ link: { block }, children }: HiddenIfInvalidLinkProps): React.ReactElement | null {
```
johnnyomair, line 10: Nit: Please remove the explicit return type here, since this can be inferred.

johnnyomair (approved):

dkarnutsch (approved):

PR #238, opened 2024-05-16
feat(site): add theme

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-431


site/src/theme.common.ts
```diff
@@ -0,0 +1,70 @@
...
+        fab: number;
+        speedDial: number;
+        appBar: number;
+        drawer: number;
+        modal: number;
+        snackbar: number;
+        tooltip: number;
+    };
+    typography: Record<string, FlattenInterpolation<ThemeProps<DefaultTheme>>>;
+}
```
jamesricky, line 65: I don't think we really need any these interfaces.
The types can just be inferred.

site/src/theme.starter.ts
```diff
@@ -0,0 +1,204 @@
...
+    },
+    zIndex: {
+        mobileStepper: 1000,
+        fab: 1050,
+        speedDial: 1050,
+        appBar: 1100,
+        drawer: 1200,
+        modal: 1300,
+        snackbar: 1400,
+        tooltip: 1500,
```
jamesricky, line 82: These values seem to be for very specific use cases that we probably won't have on normal websites that often...

site/src/util/spacing.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    ${({ theme }) => theme.breakpoints.LG.mediaQuery} {
+        :root {
+            --spacing-400: 172px;
+            --spacing-300: 100px;
+            --spacing-200: 72px;
+            --spacing-100: 48px;
+        }
+    }
+`;
```
jamesricky, line 56: Please add the `d` prefix to the dynamic variable names. Like the `s` for the static variables.

site/src/theme.starter.ts
```diff
@@ -0,0 +1,194 @@
...
+import { Breakpoint, ThemeInterface } from "./theme.common";
+
+const createBreakpoint = (value: number): Breakpoint => {
+    return {
+        mediaQuery: `@media (min-width: ${value}px)`,
+        value: value,
+    };
+};
+
+const themeStarter: ThemeInterface = {
```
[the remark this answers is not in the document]
SebiVPS: done

site/src/theme.starter.ts
```diff
@@ -0,0 +1,194 @@
...
+    name: "starter",
```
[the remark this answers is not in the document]
SebiVPS: removed

site/src/theme.starter.ts
```diff
@@ -0,0 +1,194 @@
...
+        primary: "Poppins, sans-serif",
+    },
+    shape: {
+        borderRadius: 4,
+    },
+    breakpoints: {
+        XS: createBreakpoint(600),
+        SM: createBreakpoint(900),
+        MD: createBreakpoint(1200),
+        LG: createBreakpoint(1600),
```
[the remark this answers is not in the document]
SebiVPS: renamed

site/src/theme.starter.ts
```diff
@@ -0,0 +1,194 @@
...
+    },
+    typography: {
```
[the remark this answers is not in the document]
SebiVPS: removed

site/src/theme.common.ts
SebiVPS: the ThemeInterface is used below to declare module.
The Interfaces make it much readable, instead of write the PaletteColor 5 times.

site/src/util/spacing.ts
SebiVPS: done

site/src/theme.starter.ts
SebiVPS: What do you suggest? 
I'm also absolutely no fan of this, IMO this should be placed into a utils helper file or something similar.
I just added it because the ticket explicitly says so. (You defined it in the styling focus group?)

site/src/theme.common.ts
jamesricky: > the ThemeInterface is used below to declare module.

Can't you just use the inferred type?

> The Interfaces make it much more readable, instead of write the PaletteColorInterface 5 times.

I don't think you need any of the interfaces, so you wouldn't have to write the `PaletteColorInterface` even once.

jamesricky: The benefit of not having fixed interfaces is, you can just add/remove values form the theme, without having to do the same change in the interface as well. 
Also less code to maintain and a file less.

jamesricky: > the ThemeInterface is used below to declare module.

You can do this instead:

```ts
declare module "styled-components" {
    type Theme = typeof theme;

    // eslint-disable-next-line @typescript-eslint/no-empty-interface
    export interface DefaultTheme extends Theme {}
}
```

Then you could also add `as const` to the `theme`, then the values can be seen in the IDEs type-preview:

Screenshot 2024-05-16 at 13 51 07

SebiVPS: That would be ofc a improvement when we want to support only one theme. 
For the use cases, when we need multiple themes, a interface would be necessary / recommended.

SebiVPS: i removed the interfaces now, so the starter supports one theme

site/src/theme.starter.ts
SebiVPS: removed it for now, needs to be discussed and can be added later

site/src/theme.common.ts
jamesricky: You could do something like this to have multiple themes with the same type:

```ts
const theme1 = {
    foo: "bar",
};

const theme2: typeof theme1 = {
    foo: "bar2",
};
```

site/src/theme.ts
```diff
@@ -0,0 +1,74 @@
...
+        success: {
+            light: "#77d06a",
+            main: "#66c54f",
+            dark: "#488938",
+            contrastText: "#090909",
+        },
+        common: {
+            black: "#000000",
+            white: "#ffffff",
+        },
```
jamesricky, line 55: I think this is unnecessary

```suggestion
```

site/src/theme.ts
```diff
@@ -0,0 +1,74 @@
...
+            main: "#5cb4f0",
+            dark: "#417da6",
+            contrastText: "#090909",
+        },
+        success: {
+            light: "#77d06a",
+            main: "#66c54f",
+            dark: "#488938",
+            contrastText: "#090909",
+        },
```
jamesricky, line 51: The task states the following colors should be defined: `primary`, `grey`, `text`, `success`, `error`
This is different 🤔

SebiVPS: This values come from the figma design 🤔

site/src/theme.ts
SebiVPS: I think its very nice to have them, instead of write the hex code 🤔 (shouldn't we avoid hardcoded colors?)
they were also defined in the figma design guideline

jamesricky: I don't think anyone would use a variable, when they explicitly want black/white (at least I've never done this).
It's way easier/faster to just write `color: white`. 

For content shown inside elements with a theme color as the background you would use that colors `contrastText` anyway.

site/src/theme.ts
SebiVPS: added text color 👍

site/src/theme.ts
```diff
@@ -0,0 +1,74 @@
...
+        sm: createBreakpoint(900),
+        md: createBreakpoint(1200),
+        lg: createBreakpoint(1600),
+    },
+};
+
+declare module "styled-components" {
+    type ThemeInterface = typeof theme;
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    export interface DefaultTheme extends ThemeInterface {}
```
jamesricky, line 71: Technically this is a type, not an interface 😄 

```suggestion
    type Theme = typeof theme;
    // eslint-disable-next-line @typescript-eslint/no-empty-interface
    export interface DefaultTheme extends Theme {}
```

jamesricky (dismissed):

jamesricky (approved):

PR #239, opened 2024-05-16
Add pull request template similar to comet

jamesricky:


dkarnutsch:
This file needs to be removed when using @comet/create-app.

jamesricky:
> This file needs to be removed when using @comet/create-app.

The whole `.github` directory is already removed here: https://github.com/vivid-planet/comet-starter/blob/main/create-app/src/scripts/create-app/cleanupWorkingDirectory.ts#L5C52-L5C58

dkarnutsch (approved):

PR #241, opened 2024-05-16
Add typography and use it for rich text block

SebiVPS:
Add `Typography` component and use it for `RichTextBlock`.

---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-715

<details>
Screenshot 2024-05-29 at 08 52 49
</details>


site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -117,15 +122,3 @@ export const RichTextBlock = withPreview(
...
-const Text = styled.p`
-    white-space: pre-line;
-
-    // Workaround when empty paragraphs are used as "spacing" in content
-    &:empty {
-        :before {
-            white-space: pre;
-            content: " ";
-        }
-    }
```
jamesricky, line 130: We probably still need this workaround. 
Maybe in the `Typography` component directly 🤔

site/src/components/common/TypographyStyle.ts
```diff
@@ -0,0 +1,125 @@
...
+            font-size: 14px;
+            line-height: 120%;
+            font-weight: 300;
+            margin-bottom: 10px;
+        `,
+        buttonText: css`
+            font-size: 16px;
+            line-height: 110%;
+            font-weight: 600;
+        `,
```
jamesricky, line 121: I don't think we need this, as the button is a separate component that will style itself, if the styling differs from an existing typography variant.

site/src/components/common/TypographyStyle.ts
```diff
@@ -0,0 +1,125 @@
+import { css } from "styled-components";
+
+const TypographyStyle = {
+    fontFamily: "Helvetica, sans-serif",
```
jamesricky, line 4: I think the font-family should be defined in the theme 🤔

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export interface TypographyProps extends React.HTMLAttributes<HTMLElement> {
+    component?: keyof HTMLElementTagNameMap;
+    variant: TypographyVariant;
+    disableMargin?: boolean;
+    children?: React.ReactNode;
+}
+
+export const Typography = ({ component = "div", variant, disableMargin, children, ...restProps }: TypographyProps): React.ReactElement => (
+    <Text component={component} disableMargin={disableMargin} variant={variant} as={component} {...restProps}>
```
jamesricky, line 15: I think the component/as should be the corresponding headline-/p-tag, depending on the variant. 
You could add a `variantToElementMapping` object. 

Also I think you only need `as` not `component` as well.

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,25 @@
+import { TypographyStyle } from "@src/components/common/TypographyStyle";
+import * as React from "react";
+import styled from "styled-components";
+
+export type TypographyVariant = "h600" | "h550" | "h500" | "h450" | "h350" | "h400" | "h300" | "p200" | "p100" | "buttonText";
+
+export interface TypographyProps extends React.HTMLAttributes<HTMLElement> {
+    component?: keyof HTMLElementTagNameMap;
+    variant: TypographyVariant;
+    disableMargin?: boolean;
```
jamesricky, line 10: The margin should not be there by default, we usually don't want it. 
I think we should do this like MUI's Typography, which has an optional `gutterBottom` prop, which sets the margin, when set.

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,25 @@
+import { TypographyStyle } from "@src/components/common/TypographyStyle";
+import * as React from "react";
+import styled from "styled-components";
+
+export type TypographyVariant = "h600" | "h550" | "h500" | "h450" | "h350" | "h400" | "h300" | "p200" | "p100" | "buttonText";
+
+export interface TypographyProps extends React.HTMLAttributes<HTMLElement> {
+    component?: keyof HTMLElementTagNameMap;
+    variant: TypographyVariant;
```
jamesricky, line 9: I think the `variant` should be optional and have the most common paragraph-variant as the default value, probably `p200`.

site/src/components/common/TypographyStyle.ts
```diff
@@ -0,0 +1,125 @@
+import { css } from "styled-components";
+
+const TypographyStyle = {
```
jamesricky, line 3: Should this just be defined in the component's file? 
Also, this should probably be typed to make sure every key from `TypographyVariant` is defined.

site/src/components/common/Typography.tsx
SebiVPS: It was explicitly requested that the variant is not mapped to the component tag. (eg. the headline 1 styling should also be available for h2 tags ...)
Maybe we can improve the mapping which is done right now in the draft renderer 🤔

site/src/components/common/Typography.tsx
SebiVPS: so there is no default margin? 
For the dynamic styling spaces it would makes sense that there is a default dynamic margin bottom for each breakpoint, which describes the good looking space between typography variants in continuous text. I think it would be more complex to handle this default space all around the project then disable it, when its not used with this flag.
(the flag is used in all projects i saw right now, and it makes sense for me, but i have no better solution right now)

site/src/components/common/TypographyStyle.ts
SebiVPS: When we should place it into the Typography file itself, it could not be generated 🤔 
Typing is a good idea 👍

site/src/components/common/Typography.tsx
SebiVPS: i added the gutterBottom prop as discussed 👍

site/src/components/common/TypographyStyle.ts
SebiVPS: i moved the style mapping into the component's file, @MichaelSchwaiger48 can review the auto generating of the styling separately

site/src/components/common/Typography.tsx
SebiVPS: @jamesricky at the moment i added a fixed space of Spacing.S300 for all text elements when gutterBottom is set. When UX decided how to handle this in the future we can implement this here.

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    white-space: pre-line;
+
+    // Workaround when empty paragraphs are used as "spacing" in content
+    &:empty {
+        :before {
+            white-space: pre;
+            content: " ";
+        }
+    }
```
SebiVPS, line 134: @jamesricky do we really need this, seems to be a pretty special edge case. Or is this common that an empty rte is used as spacing?
In my opinion the user should use the space blocks instead of an empty rte block?

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,134 @@
...
+export const Typography = ({ component = "div", children, ...restProps }: TypographyProps): ReactElement => (
+    <Text as={component} {...restProps}>
+        {children}
+    </Text>
+);
+
+const Text = styled.div<TypographyProps>`
+    font-family: ${({ theme }) => theme.fontFamily};
+    color: ${({ theme, color }) => (color ? color : theme.palette.text.primary)};
+    ${({ variant = "p200" }) => TypographyVariantStyle[variant]};
```
jamesricky, line 121: I would prefer the default variant be set in the `Typography` component, instead of the styled-component. 
It would make it easer to use the variant if it's needed somewhere else in the future.

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export const Typography = ({ component = "div", children, ...restProps }: TypographyProps): ReactElement => (
+    <Text as={component} {...restProps}>
+        {children}
+    </Text>
+);
+
+const Text = styled.div<TypographyProps>`
+    font-family: ${({ theme }) => theme.fontFamily};
+    color: ${({ theme, color }) => (color ? color : theme.palette.text.primary)};
```
jamesricky, line 120:
```suggestion
    color: ${({ theme, color }) => (color || theme.palette.text.primary)};
```

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export interface TypographyProps extends HTMLAttributes<HTMLElement> {
+    component?: keyof HTMLElementTagNameMap;
+    variant?: TypographyVariant;
+    color?: string;
+    gutterBottom?: boolean;
+    children?: ReactNode;
+}
+
+export const Typography = ({ component = "div", children, ...restProps }: TypographyProps): ReactElement => (
```
jamesricky, line 112: The return type can just be inferred

```suggestion
export const Typography = ({ component = "div", children, ...restProps }: TypographyProps) => (
```

site/src/theme.ts
```diff
@@ -30,12 +30,25 @@ export const theme = {
             900: "#212121",
         },
     },
+    fontFamily: "Arial",
```
jamesricky, line 33: Please add a fallback font

```suggestion
    fontFamily: "Arial, sans-serif",
```

site/src/components/common/Typography.tsx
jamesricky: > It was explicitly requested that the variant is not mapped to the component tag.

By who? Why?

This would require setting a component for every single usage of `Typography`...
I think if there is no explicit `component` set, there should be a default, depending on the variant. 

If not, we have to make the `component` prop required and remove the default `div` otherwise this will be forgotten 90% of the time, which is likely bad for SEO and accessibility .

site/src/components/common/Typography.tsx
SebiVPS: ok, changed it 👍

site/src/components/common/Typography.tsx
SebiVPS: > By who? Why?
@thomasdax98

SebiVPS: I copied the component from the latest project, how do you use it there?

jamesricky: > Maybe a span tag does make more sense than a div as default?

No, text should be in a `p` or headline tag. 

If there is any default, it should be `p` but I think it should depend on the variant.

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const Text = styled.div<TextProps>`
+    font-family: ${({ theme }) => theme.fontFamily};
+    color: ${({ theme, color }) => color || theme.palette.text.primary};
+    ${({ variant }) => typographyVariantStyle[variant]};
+    ${({ theme, gutterBottom }) =>
+        gutterBottom &&
+        css`
+            margin-bottom: ${theme.spacing.S300};
+        `}
```
jamesricky, line 132: Some tags have a default bottom margin, that should be removed, if it's not overridden. 

```suggestion
    margin-bottom: ${({ theme, gutterBottom }) => (gutterBottom ? theme.spacing.S300 : 0)};
```

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -118,14 +118,10 @@ export const RichTextBlock = withPreview(
     { label: "Rich Text" },
 );
 
-const Text = styled.p`
-    white-space: pre-line;
```
jamesricky, line 122: I think we should not remove this 🤔

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -118,14 +118,10 @@ export const RichTextBlock = withPreview(
...
-
-    // Workaround when empty paragraphs are used as "spacing" in content
-    &:empty {
-        :before {
-            white-space: pre;
-            content: " ";
-        }
+const StyledTyphography = styled(Typography)`
```
jamesricky, line 121: Should the name maybe stay `Text`? 
`Styled` doesn't really make sense, as `Typography` is already styled.

jamesricky, line 121: Should this not be used for all blocks?

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -27,61 +28,60 @@ const defaultRenderers: Renderers = {
      */
     blocks: {
         // Paragraph
-        unstyled: (children, { keys }) => children.map((child, idx) => <Text key={keys[idx]}>{child}</Text>),
+        unstyled: (children, { keys }) =>
```
jamesricky, line 31: Is `unstyled` really the best option here? 
Should this maybe be `paragraph` or `paragraph-one` to match the naming of the headlines?
Then you could set that as the `standardBlockType`. 

Also the comments `// Paragraph`, `// Headlines` can be removed, that's already obvious from the code.

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -27,61 +28,60 @@ const defaultRenderers: Renderers = {
...
+            children.map((child, idx) => (
+                <StyledTyphography component={"p"} key={keys[idx]} gutterBottom>
```
jamesricky, line 33: I think it would be cleaner to remove the `{}` from all the `variant` and `component` props. 

```suggestion
                <StyledTyphography component="p" key={keys[idx]} gutterBottom>
```

site/src/components/common/Typography.tsx
jamesricky: I just spoke with @thomasdax98, he agrees it makes sense to have a variant-to-tag mapping for the default value.

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: I thought it would only makes sense for paragraph elements? I do not see the use case for a headline component, what do you think?

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: There would be a linter rule, to automatically change / force that. 
I would suggest to add the linter rule to use it consistent project wide, but that would be an own PR.
@thomasdax98 should we add that rule?
Is that what we are looking for [jsx-curly-brace-presence](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md)

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: I did not took a closer look to the rte implementation yet, just implemented the typography component.
In the api the standard implementation with `createRichTextBlock` with no additional config is used atm. 
The second paragraph setting (p200, p100) from the design is also still missing. I thought about an own PR for the RTE changes.

site/src/components/common/Typography.tsx
SebiVPS: good idea 👍  the default mapping is added, fallback is a p tag now

nsams:
please rebase, this PR should not contain commits from https://github.com/vivid-planet/comet-starter/pull/244

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: ok added it for all blocks

admin/src/common/blocks/RichTextBlock.tsx
```diff
@@ -1,5 +1,42 @@
...
+        ],
+        standardBlockType: "paragraph-one",
+        blocktypeMap: {
+            "paragraph-one": {
+                label: <FormattedMessage id="richTextBlock.paragraph-one" defaultMessage="Paragraph 1" />,
+            },
+            "paragraph-two": {
+                label: <FormattedMessage id="richTextBlock.paragraph-two" defaultMessage="Paragraph 2" />,
+                renderConfig: {
+                    element: (p) => <Typography paragraph fontSize={14} {...p} />,
```
jamesricky, line 37: MUIs Typography already has a variant for a smaller body text. 

```suggestion
                    element: (p) => <Typography paragraph variant="body2" {...p} />,
```

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: added necessary changes to the RTE Block itself 👍

admin/src/common/blocks/RichTextBlock.tsx
```diff
@@ -1,5 +1,42 @@
...
+            "header-three",
+            "header-four",
+            "header-five",
+            "ordered-list",
+            "unordered-list",
+            "link",
+            "links-remove",
+            "non-breaking-space",
+            "soft-hyphen",
+        ],
```
jamesricky, line 28: @johnnyomair @thomasdax98 

`supports` only was overridden, because we need to remove `header-six` from the default blocks, as this does not exist according to our designs. I don't think this should be necessary, as this is our default. 

Would it make sense to remove `header-six` from `createRichTextBlock()` or would this be too much of a breaking change for existing projects?

Or should we ask UX to add another variant for the headline?

admin/src/common/blocks/RichTextBlock.tsx
```diff
@@ -1,5 +1,42 @@
...
+        standardBlockType: "paragraph-one",
+        blocktypeMap: {
+            "paragraph-one": {
+                label: <FormattedMessage id="richTextBlock.paragraph-one" defaultMessage="Paragraph 1" />,
+            },
+            "paragraph-two": {
```
jamesricky, line 34: @johnnyomair @thomasdax98 

Should a value for `standardBlockType`, other than `unstyled` be included in `createRichTextBlock()` by default?
Also, does it make sense to include two paragraphs by default in `blocktypeMap` of `createRichTextBlock()`?

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -26,42 +27,45 @@ const defaultRenderers: Renderers = {
...
+                <Text key={keys[idx]} gutterBottom>
+                    {child}
+                </Text>
+            )),
+        "paragraph-two": (children, { keys }) =>
+            children.map((child, idx) => (
+                <Text variant="p100" key={keys[idx]} gutterBottom>
+                    {child}
+                </Text>
+            )),
```
jamesricky, line 41: It seems a bit weird to have `paragraph-one` styled as `p200` and `paragraph-two` as `p100` 🤔 
Is this correct?

SebiVPS: funnily enough it is :) a bit unlucky with the values 😅 
I used the existing  convention (header-one, header two, ... where h600 is the largest) and adapt it for the paragraphs. 
Would you suggest a change? If we name it p200 and p100 it would be clearer, but then it does not fit with the current headline conventions.

admin/src/common/blocks/RichTextBlock.tsx
johnnyomair: Maybe we should just add a sixth headline. I don't understand why we have h1-h5, but no h6 for HeadingBlock either 🤷🏼‍♂️

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: @johnnyomair how can this be a breaking change? This rule should only apply to the code style in this repo and has no effect on interfaces? Or do i miss something?

admin/src/common/blocks/RichTextBlock.tsx
SebiVPS: Ok, i'll ask Mitch for an additional styling for h6. I guess the UX team wants to simplify the usage, i also think that 5 headline variants can be enough.

admin/src/common/blocks/RichTextBlock.tsx
johnnyomair: > Should a value for standardBlockType, other than unstyled be included in createRichTextBlock() by default?

What would be the benefit of this? For some rich texts, e.g., an eyebrow with online inline styling, having one block type `unstyled` is perfectly fine. Also unstyled is widely used in the RTE as default, I believe changing it only for the rich text block may be confusing.

> Also, does it make sense to include two paragraphs by default in blocktypeMap of createRichTextBlock()?

Based on the above mentioned example of the eyebrow, I'd say no.

admin/src/common/blocks/RichTextBlock.tsx
```diff
@@ -1,5 +1,42 @@
...
+                label: <FormattedMessage id="richTextBlock.paragraph-two" defaultMessage="Paragraph 2" />,
+                renderConfig: {
+                    element: (p) => <Typography paragraph variant="body2" {...p} />,
```
johnnyomair, line 37:
```suggestion
                    element: (props) => <Typography paragraph variant="body2" {...props} />,
```

Please don't use abbreviations

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: I changed the nameing as discussed :)

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -26,42 +27,57 @@ const defaultRenderers: Renderers = {
...
+                <Text key={keys[index]} gutterBottom>
+                    {child}
+                </Text>
+            )),
+        "paragraph-small": (children, { keys }) =>
+            children.map((child, index) => (
+                <Text variant="p100" key={keys[index]} gutterBottom>
+                    {child}
+                </Text>
+            )),
```
jamesricky, line 47: The tasks says paragraphs `200` and `300` should be used. 
You are using `200`(the default) and `100`.

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -26,42 +27,57 @@ const defaultRenderers: Renderers = {
...
     blocks: {
-        // Paragraph
-        unstyled: (children, { keys }) => children.map((child, idx) => <Text key={keys[idx]}>{child}</Text>),
-        // Headlines
+        unstyled: (children, { keys }) =>
+            children.map((child, index) => (
+                <Text key={keys[index]} gutterBottom>
+                    {child}
+                </Text>
+            )),
```
jamesricky, line 35: There should be no `unstyled` when there are already two paragraph blocks available.

admin/src/common/blocks/RichTextBlock.tsx
```diff
@@ -1,5 +1,23 @@
...
+export const RichTextBlock = createRichTextBlock({
+    link: LinkBlock,
+    rte: {
+        standardBlockType: "paragraph-standard",
+        blocktypeMap: {
+            "paragraph-standard": {
+                label: <FormattedMessage id="richTextBlock.paragraph-one" defaultMessage="Paragraph Standard" />,
+            },
+            "paragraph-small": {
+                label: <FormattedMessage id="richTextBlock.paragraph-two" defaultMessage="Paragraph Small" />,
```
jamesricky, line 16: The IDs are incorrect and should be camelCase. 

```suggestion
                label: <FormattedMessage id="richTextBlock.paragraphStandard" defaultMessage="Paragraph Standard" />,
            },
            "paragraph-small": {
                label: <FormattedMessage id="richTextBlock.paragraphSmall" defaultMessage="Paragraph Small" />,
```

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: The values were defined in the design.
Should i change the ticket, or should i talk to Mitch to adapt the design?

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: I talked with Mario and Obi, we came to the result that the standard rte block is often used for other implementations wich do not define paragraphs, so a unstyled fallback here would make sense, even if its not used for the standard block itself.

site/src/common/blocks/RichTextBlock.tsx
jamesricky: > The values were defined in the design.
> Should i change the ticket, or should i talk to Mitch to adapt the design?

It's probably fine, if that's how it's defined in the design 👍🏼 
_(Can't hurt to ask though, just to be sure)_ 

> For the headlines there is also no h350 in the ticket 👀

True, but we decided to add this so we don't have to override a bunch of settings, just to remove `headline-six`.

site/src/common/blocks/RichTextBlock.tsx
jamesricky: Ok, but having two blocks with the exact same style probably doesn't make sense imo. 
Should we remove `paragraph-standard` then? 🤔

jamesricky (approved):

mariokemetinger (approved):

johnnyomair (approved):

PR #242, opened 2024-05-16
Add heading block

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-771

<details>
Screenshot 2024-05-24 at 08 15 11
</details>


api/src/common/blocks/heading.block.ts
```diff
@@ -0,0 +1,63 @@
...
+    H2 = "h2",
+    H3 = "h3",
+    H4 = "h4",
+    H5 = "h5",
+}
+
+export enum HeadlineTextAlignment {
+    Left = "left",
+    Center = "center",
+}
```
jamesricky, line 27: According to our coding guidelines, the key and value of enums should always be `PascalCase`:

image

When the value needed in the frontend differs from the value the API provides, you can create a mapping object for that.

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,17 @@
+import { PropsWithData, withPreview } from "@comet/cms-site";
+import { HeadingBlockData } from "@src/blocks.generated";
+import * as React from "react";
+
+import { RichTextBlock } from "./RichTextBlock";
+
+export const HeadingBlock = withPreview(
+    ({ data: { eyebrow, headline, htmlTag, textAlignment } }: PropsWithData<HeadingBlockData>) => {
+        return (
+            <div style={{ textAlign: textAlignment }}>
```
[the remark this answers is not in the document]
SebiVPS: i removed the inline style add further pointed this PR into a feature branch, since the site component is not yet implemented. (added a todo ;) )

admin/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,95 @@
+"use client";
+import { IBlocktypeMap } from "@comet/admin-rte/lib/core/types";
```
nsams, line 2: this is a private export you are not supposed to use

api/src/common/blocks/heading.block.ts
```diff
@@ -0,0 +1,63 @@
...
+
+class HeadingBlockData extends BlockData {
+    @ChildBlock(RichTextBlock)
+    eyebrow: BlockDataInterface;
+
+    @ChildBlock(RichTextBlock)
+    headline: BlockDataInterface;
+
+    @BlockField({ type: "enum", enum: HeadlineTag })
+    htmlTag: HeadlineTag;
```
johnnyomair, line 37: We're developing a headless CMS, so using a name such as htmlTag isn't the best option. For instance, a mobile app most likely won't use HTML. How about naming it `level` instead?

SebiVPS: hmm, since the prop is explicitly used to set the right HTML tag in Web and does nothing on mobile apps (the styling is applied from the RTE style prop), in my opinion another name would just be misleading. Who can we add to this dicussion?

admin/src/documents/pages/blocks/PageContentBlock.tsx
```diff
@@ -10,6 +11,7 @@ export const PageContentBlock = createBlocksBlock({
     supportedBlocks: {
         space: SpaceBlock,
         richtext: RichTextBlock,
+        heading: HeadingBlock,
         headline: HeadlineBlock,
```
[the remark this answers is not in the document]
SebiVPS: yes ofc, my plan is to remove it, when the site implementation is ready.

admin/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,92 @@
...
+        supports: ["bold", "italic", "sub", "sup", "non-breaking-space", "soft-hyphen"],
+    },
+    minHeight: 0,
+});
+
+const HeadlineRichTextBlock = createRichTextBlock({
+    link: LinkBlock,
+    rte: {
+        maxBlocks: 1,
+        supports: ["bold", "italic", "sub", "sup", "non-breaking-space", "soft-hyphen"],
```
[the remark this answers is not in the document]
SebiVPS: Isn't it then confusing if the same keys in the normal RTE, e.g. for header-one, imply that an H1 is used? Or should i just use the keys and also override the blockTypeMap with the existing labels?

admin/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,92 @@
...
+        blocktypeMap: {
+            unstyled: {
+                label: <FormattedMessage id="headingBlock.headline.size.600" defaultMessage="Size 600" />,
+            },
```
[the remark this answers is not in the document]
SebiVPS: Did not know that one before, makes absolutely sense 👍

admin/src/common/blocks/HeadingBlock.tsx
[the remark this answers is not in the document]
johnnyomair: > Or should i just use the keys and also override the blockTypeMap with the labels below?

Sounds good to me.

api/src/common/blocks/heading.block.ts
johnnyomair: @nsams what do you think?

admin/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,99 @@
...
+    link: LinkBlock,
+    rte: {
+        maxBlocks: 1,
+        supports: [
+            "header-one",
+            "header-two",
+            "header-three",
+            "header-four",
+            "header-five",
+            "bold",
```
SebiVPS, line 28: fyi: bold does not makes sense in this case, because the current font (Arial) does only support font-weight of 400 and 700 so the strong tag can not be adapted in the site. 
It is requested to support bold so we add it, it can be implemented when other fonts are used.

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+import { defaultRTERenderers, RichTextBlock } from "./RichTextBlock";
+
+// Custom renderers to render unstyled content with h400 styling for eyebrow
+const eyebrowRenderers: Renderers = {
+    inline: defaultRTERenderers.inline,
+    blocks: {
+        unstyled: (children, { keys }) =>
+            children.map((child, idx) => (
+                <Typography variant={"h400"} component={"h5"} key={keys[idx]} gutterBottom>
```
SebiVPS, line 15: The styling h400 and to use h5 tag is from the figma design.
Screenshot 2024-05-24 at 08 34 39

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -26,42 +26,45 @@ const defaultRenderers: Renderers = {
      * Note that children are an array of blocks with same styling,
      */
     blocks: {
-        // Paragraph
-        unstyled: (children, { keys }) => children.map((child, idx) => <Text key={keys[idx]}>{child}</Text>),
```
johnnyomair, line 30: I don't think you can remove that

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -9,7 +9,7 @@ import { LinkBlock } from "./LinkBlock";
 /**
  * Define the renderers
  */
-const defaultRenderers: Renderers = {
+export const defaultRTERenderers: Renderers = {
```
johnnyomair, line 12: This should either be `defaultRichTextRenderers` or simply `defaultRenderers`. No editing here

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,140 @@
...
+    h400: "h5",
+    p200: "p",
+    p100: "p",
+};
+
+export interface TypographyProps extends HTMLAttributes<HTMLElement> {
+    component?: keyof HTMLElementTagNameMap;
+    variant?: TypographyVariant;
+    color?: string;
+    gutterBottom?: boolean;
```
johnnyomair, line 118: Shouldn't this gutter be depending on the variant?

site/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,140 @@
...
+    h450: "h4",
+    h400: "h5",
+    p200: "p",
+    p100: "p",
+};
+
+export interface TypographyProps extends HTMLAttributes<HTMLElement> {
+    component?: keyof HTMLElementTagNameMap;
+    variant?: TypographyVariant;
+    color?: string;
```
johnnyomair, line 117: Does this allow setting any HTML color? Is this what we want?

admin/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,99 @@
+"use client";
```
johnnyomair, line 1:
```suggestion
```

This will not be needed in Admin for a long time 😁

site/src/components/common/Typography.tsx
SebiVPS: Yes, it should, in the future.
I discussed this with @jamesricky and Mitch. Mitch will talk with his team to decide wich value should be used for the bottom margin.

site/src/components/common/Typography.tsx
SebiVPS: Hmm, do we need to set the text color as prop at all?

site/src/common/blocks/RichTextBlock.tsx
SebiVPS: Can you explain why? I only allowed a fixed set with a default value, so there should not be an unstyled option?

site/src/components/common/Typography.tsx
SebiVPS: i removed the color prop in [!241](https://github.com/vivid-planet/comet-starter/pull/241/commits/78abda0f26c93c912b10966407e6b3806d30ef8e) since that belongs to this PR.
If there is the need to set the color, it can be added.

site/src/components/common/Typography.tsx
jamesricky: For now I think it's fine to have a fixed value.

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,94 @@
...
+    inline: defaultRichTextRenderers.inline,
+    blocks: {
+        unstyled: (children, { keys }) =>
+            children.map((child, index) => (
+                <Typography variant={"h400"} component={"h5"} key={keys[index]} gutterBottom>
+                    {child}
+                </Typography>
+            )),
+    },
+};
```
thomasdax98, line 20: @mariokemetinger @jamesricky Should the Typography be here in the renderer or as a wrapper in `HeadingBlock`?

admin/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,103 @@
...
+                label: <FormattedMessage id="headingBlock.headline.size.500" defaultMessage="Size 500" />,
+            },
+            "header-four": {
+                label: <FormattedMessage id="headingBlock.headline.size.450" defaultMessage="Size 450" />,
+            },
+            "header-five": {
+                label: <FormattedMessage id="headingBlock.headline.size.400" defaultMessage="Size 400" />,
+            },
+            "header-six": {
+                label: <FormattedMessage id="headingBlock.headline.size.350" defaultMessage="Size 350" />,
```
jamesricky, line 53: The IDs are not really consistent with the rest of the file, maybe change it to something like this: 

```suggestion
                label: <FormattedMessage id="headingBlock.size600" defaultMessage="Size 600" />,
            },
            "header-two": {
                label: <FormattedMessage id="headingBlock.size550" defaultMessage="Size 550" />,
            },
            "header-three": {
                label: <FormattedMessage id="headingBlock.size500" defaultMessage="Size 500" />,
            },
            "header-four": {
                label: <FormattedMessage id="headingBlock.size450" defaultMessage="Size 450" />,
            },
            "header-five": {
                label: <FormattedMessage id="headingBlock.size400" defaultMessage="Size 400" />,
            },
            "header-six": {
                label: <FormattedMessage id="headingBlock.size350" defaultMessage="Size 350" />,
```

site/src/common/blocks/HeadingBlock.tsx
jamesricky: @SebiVPS and I discussed this a few days ago.

Generally it makes sense to render the `Typography` inside the renderer to allow multiple paragraphs/blocks, each with their own `gutterBottom` from `Typography`. 

In this specific case however, multiple paragraphs/blocks are not supported (due to `maxBlocks: 1` in admin) so it's not really necessary inside the renderer, also having a styled text inside something called `unstyled` is a bit confusing 😅 

So here it probably makes sense to move the `Typography` to the component and keep this actually "unstyled".

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,94 @@
...
+const headlineTagMap: Record<HeadingBlockData["htmlTag"], keyof HTMLElementTagNameMap> = {
+    H1: "h1",
+    H2: "h2",
+    H3: "h3",
+    H4: "h4",
+    H5: "h5",
+    H6: "h6",
+};
+
+const textAlignmentMap: Record<HeadingBlockData["textAlignment"], "left" | "center"> = {
```
jamesricky, line 74: You could use the `textAlign` type here. 
Then there's one less thing to change when changing options in the future. 

```suggestion
const textAlignmentMap: Record<HeadingBlockData["textAlignment"], React.CSSProperties["textAlign"]> = {
```

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,94 @@
...
+            <Root $textAlign={textAlignmentMap[textAlignment]}>
+                <RichTextBlock data={eyebrow} renderers={eyebrowRenderers} />
+                <RichTextBlock data={headline} renderers={getHeadlineRenderers(headlineTag)} />
+            </Root>
+        );
+    },
+    { label: "Headline" },
+);
+
+const Root = styled.div<{ $textAlign: "left" | "center" }>`
```
jamesricky, line 92:
```suggestion
const Root = styled.div<{ $textAlign: React.CSSProperties["textAlign"] }>`
```

site/src/common/blocks/HeadingBlock.tsx
SebiVPS: looks better, when the typography is wrapped outside the rte 👍

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,87 @@
...
+        return (
+            <Root $textAlign={textAlignmentMap[textAlignment]}>
+                <Typography variant={"h400"} component={"h5"} gutterBottom>
+                    <RichTextBlock data={eyebrow} renderers={eyebrowRenderers} />
+                </Typography>
+                <RichTextBlock data={headline} renderers={getHeadlineRenderers(headlineTag)} />
+            </Root>
+        );
+    },
+    { label: "Headline" },
```
johnnyomair, line 82: Shouldn't this be "Heading"?

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -0,0 +1,87 @@
...
+const textAlignmentMap: Record<HeadingBlockData["textAlignment"], CSSProperties["textAlign"]> = {
+    Left: "left",
+    Center: "center",
+};
+
+export const HeadingBlock = withPreview(
+    ({ data: { eyebrow, headline, htmlTag, textAlignment } }: PropsWithData<HeadingBlockData>) => {
+        const headlineTag = headlineTagMap[htmlTag];
+        return (
+            <Root $textAlign={textAlignmentMap[textAlignment]}>
```
johnnyomair, line 74: @jamesricky should we use the `OwnerState` pattern for single props as well?

site/src/common/blocks/HeadingBlock.tsx
SebiVPS: indeed 👍

johnnyomair (approved):

jamesricky (approved):

mariokemetinger (approved):

PR #244, opened 2024-05-17
Add site theme

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-431

### DIFF
--- a/site/src/theme.ts
+++ b/site/src/theme.ts
@@ -5,7 +5,7 @@ const createBreakpoint = (value: number) => {
     };
 };
 
-const theme = {
+export const theme = {
     palette: {
         primary: {
             light: "#4b9fea",
@@ -62,5 +62,3 @@ declare module "styled-components" {
     // eslint-disable-next-line @typescript-eslint/no-empty-interface
     export interface DefaultTheme extends Theme {}
 }
-
-export { theme };
--- a/site/src/theme.ts
+++ b/site/src/theme.ts
@@ -1,0 +1,74 @@
+const createBreakpoint = (value: number) => {
+    return {
+        mediaQuery: `@media (min-width: ${value}px)`,
+        value: value,
+    };
+};
+
+const theme = {
+    palette: {
+        primary: {
+            light: "#4b9fea",
+            main: "#1e88e5",
+            dark: "#155fa0",
+            contrastText: "#ffffff",
+        },
+        text: {
+            primary: "#212121",
+            secondary: "#757575",
+        },
+        grey: {
+            50: "#fafafa",
+            100: "#f5f5f5",
+            200: "#eeeeee",
+            300: "#e0e0e0",
+            400: "#bdbdbd",
+            500: "#9e9e9e",
+            600: "#757575",
+            700: "#616161",
+            800: "#424242",
+            900: "#212121",
+        },
+        error: {
+            light: "#c6523f",
+            main: "#bc3520",
+            dark: "#832617",
+            contrastText: "#ffffff",
+        },
+        warning: {
+            light: "#f5c15f",
+            main: "#f3b346",
+            dark: "#a87d32",
+            contrastText: "#090909",
+        },
+        info: {
+            light: "#73c1f2",
+            main: "#5cb4f0",
+            dark: "#417da6",
+            contrastText: "#090909",
+        },
+        success: {
+            light: "#77d06a",
+            main: "#66c54f",
+            dark: "#488938",
+            contrastText: "#090909",
+        },
+    },
+    shape: {
+        borderRadius: 4,
+    },
+    breakpoints: {
+        xs: createBreakpoint(600),
+        sm: createBreakpoint(900),
+        md: createBreakpoint(1200),
+        lg: createBreakpoint(1600),
+    },
+};
+
+declare module "styled-components" {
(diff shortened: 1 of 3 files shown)


nsams (dismissed):

jamesricky (approved):

jamesricky (dismissed):

site/src/theme.ts
```diff
@@ -0,0 +1,74 @@
...
+        success: {
+            light: "#77d06a",
+            main: "#66c54f",
+            dark: "#488938",
+            contrastText: "#090909",
+        },
+    },
+    shape: {
+        borderRadius: 4,
+    },
```
johnnyomair, line 59: What is this used for?

SebiVPS: the shape.borderRadius is used for all kind of components which should use a border radius according to design (buttons, cards, ...)

johnnyomair: I've discussed this with @mariokemetinger: The border radius for different elements (buttons, cards etc.) will typically be different, so having a generic shape.borderRadius would be confusing. I'd rather remove it. We may add concrete button styles to the theme (to be able to style anything like a button, e.g., a link) in the future (however, I'm still not convinced that this is actually a good idea).

SebiVPS: ok its removed 👍

site/src/theme.ts
```diff
@@ -0,0 +1,71 @@
...
+    breakpoints: {
+        xs: createBreakpoint(600),
+        sm: createBreakpoint(900),
+        md: createBreakpoint(1200),
+        lg: createBreakpoint(1600),
+    },
+};
+
+declare module "styled-components" {
+    type Theme = typeof theme;
```
johnnyomair, line 66: Should we declare this type outside of styled-components and export it? @jamesricky what do you think?

SebiVPS: I do not think it is necessary atm for the starter project, when the normal usecase is, that only one theme is needed. For multiple themes a separate theme interface would be recommended i guess.

johnnyomair: I'd still declare the `Theme` type outside of the styled-components module, even if we don't export it. Defining types in a foreign module may lead to unexpected behaviour.

SebiVPS: moved it outside 👍

johnnyomair:
LGTM, @jamesricky @mariokemetinger what do you think?

dkarnutsch (approved):

site/src/theme.ts
```diff
@@ -0,0 +1,72 @@
...
+            dark: "#832617",
+            contrastText: "#ffffff",
+        },
+        warning: {
+            light: "#f5c15f",
+            main: "#f3b346",
+            dark: "#a87d32",
+            contrastText: "#090909",
+        },
+        info: {
```
mariokemetinger, line 44: What is the use-case for "info"? Would delete that...

site/src/util/spacing.ts
```diff
@@ -0,0 +1,58 @@
...
+    D300 = "var(--spacing-d300)",
+    D200 = "var(--spacing-d200)",
+    D100 = "var(--spacing-d100)",
+    S600 = "var(--spacing-s600)",
+    S500 = "var(--spacing-s500)",
+    S400 = "var(--spacing-s400)",
+    S300 = "var(--spacing-s300)",
+    S200 = "var(--spacing-s200)",
+    S100 = "var(--spacing-s100)",
+}
```
jamesricky, line 14: This does not adhere to our enum naming conventions...
Do we actually need this to be an enum? Can't it just be an object?

```ts
const spacing = {
    d400: "var(--spacing-d400)",
    d300: "var(--spacing-d300)",
    // ...
};
```

site/src/theme.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+type Theme = typeof theme;
+
+declare module "styled-components" {
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    export interface DefaultTheme extends Theme {}
+}
+
+export { theme };
```
jamesricky, line 66: I don't think we should have a separate export line, just export the value directly.

site/src/theme.ts
SebiVPS: mitch had it in the design, Ricky and Mitch also discussed before. Mitch will leave it in the design, i removed it now, since we don't have components that use that color i guess. 
We can readd it if we need it.

site/src/theme.ts
```diff
@@ -0,0 +1,66 @@
...
+            main: "#f3b346",
+            dark: "#a87d32",
+            contrastText: "#090909",
+        },
+        success: {
+            light: "#77d06a",
+            main: "#66c54f",
+            dark: "#488938",
+            contrastText: "#090909",
+        },
```
jamesricky, line 49: @mariokemetinger since you don't think we should have the `info` color, do we really need any of these colors? The only use case I can think of here is errors in form components, which we currently do not have.

site/src/util/spacing.ts
SebiVPS: made it an object 👍

jamesricky (approved):

mariokemetinger (approved):

johnnyomair (approved):

PR #247, opened 2024-05-21
Set imgproxy image to version v3

Vierlip:
Jira: Erweiterung CLI: Update ImgProxy von v2 auf v3

Timr: INFRA-271

Update imgproxy version to use v3 instead of latest

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - postgres
 
   imgproxy:
-    image: darthsim/imgproxy:latest
+    image: darthsim/imgproxy:v3
     volumes:
       - ./api/uploads:/uploads:ro
     ports:


dkarnutsch:
**Further clarification (as this task was created by me):** Currently local env and deployed env are drifting. While we use locally always the latest version, we pin the deployed version (to v2, but that is another topic). We should also pin the local version to ensure they stay in sync.

nsams (approved):

johnnyomair (approved):

johnnyomair:
> **Further clarification (as this task was created by me):** Currently local env and deployed env are drifting. While we use locally always the latest version, we pin the deployed version (to v2, but that is another topic). We should also pin the local version to ensure they stay in sync.

Thanks for the clarification! @Vierlip this information is something that you could add to a PR's description. It's more important to state _why_ you're making a change than _what_ the change is (mostly, this can be seen in the code). Doing so would help us review the change more quickly. Thanks! ☺️

dkarnutsch (approved):

thomasdax98 (approved):

PR #249, opened 2024-05-22
feature: add call to action button (and list) block

SebiVPS:
TODO:
- [] add DamFileDownloadLinkBlock [comet !1228](https://github.com/vivid-planet/comet/pull/1228)
- [] site implementation needs [!241](https://github.com/vivid-planet/comet-starter/pull/241) and basic button component

---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-774

<details>
    <summ
Screenshot 2024-05-22 at 15 04 02
ary>Screenshots/screencasts</summary>
</details>


site/src/theme.ts
```diff
@@ -18,16 +18,16 @@ export const theme = {
...
+            _050: "#fafafa",
+            _100: "#f5f5f5",
+            _200: "#eeeeee",
+            _300: "#e0e0e0",
+            _400: "#bdbdbd",
+            _500: "#9e9e9e",
+            _600: "#757575",
+            _700: "#616161",
+            _800: "#424242",
+            _900: "#212121",
```
SebiVPS, line 30: @jamesricky @mariokemetinger @thomasdax98 
I changed the keys to _xxx to simplify the usage in css. This way we can write `theme.palette.grey._100` the other way we had to write `theme.palette.grey["100"]`. 
I prefer the current implementation to have the consistent usage and the IDE autocomplete suggestions after writing the dot.

admin/src/common/blocks/CallToActionButtonBlock.tsx
```diff
@@ -0,0 +1,34 @@
+import { BlockCategory, createCompositeBlock, createCompositeBlockSelectField } from "@comet/blocks-admin";
+import { CallToActionButtonBlockData } from "@src/blocks.generated";
+import * as React from "react";
+import { FormattedMessage } from "react-intl";
+
+import { TextLinkBlock } from "./TextLinkBlock";
+
+export const CallToActionButtonBlock = createCompositeBlock(
```
thomasdax98, line 8: CallToActionButtonBlock -> CallToActionBlock

Generally Button -> CallToAction

site/src/theme.ts
jamesricky: Good point 👍🏼 

As an alternative, we would do something like this. 
Although I generally like having the individual values grouped, if they belong together 🤔 

```ts
palette: {
    grey50: "#fafafa",
    grey100: "#f5f5f5",
    // ...
}
```

@mariokemetinger @johnnyomair @thomasdax98 what do you think?

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,101 @@
...
+    );
+};
+
+export const Root = styled.div<{ $variant: ButtonVariant; $disabled: boolean }>`
+    display: inline-flex;
+    box-sizing: border-box;
+    padding: ${({ theme }) => `${theme.spacing.S400} ${theme.spacing.S500}`};
+    border-radius: 4px;
+    cursor: pointer;
+    user-select: none;
```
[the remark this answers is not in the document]
SebiVPS: i removed it, i thought it would not be necessary to select a text in a button

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,101 @@
...
+    justify-content: center;
+    align-items: center;
+    transition: all 0.2s ease-out;
+
+    text-align: center;
+    text-decoration: none;
+    font-family: ${({ theme }) => theme.fontFamily};
+    font-size: 16px;
+    font-weight: 700;
+    line-height: 110%;
```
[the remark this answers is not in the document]
SebiVPS: no, not really, thats why we had a own buttonText Style

jamesricky (approved):

api/src/common/blocks/call-to-action.block.ts
```diff
@@ -0,0 +1,43 @@
...
+
+enum Variant {
+    Contained = "Contained",
+    Outlined = "Outlined",
+    Text = "Text",
+}
+
+class CallToActionBlockData extends BlockData {
+    @ChildBlock(TextLinkBlock)
+    textLink: BlockDataInterface;
```
johnnyomair, line 24: IMO `link` would be enough, what do you think?

api/src/documents/pages/blocks/page-content.block.ts
```diff
@@ -17,7 +18,8 @@ export const PageContentBlock = createBlocksBlock(
             textImage: TextImageBlock,
             damVideo: DamVideoBlock,
             youTubeVideo: YouTubeVideoBlock,
-            links: LinkListBlock,
+            callToAction: CallToActionBlock,
```
johnnyomair, line 21: Should it be possible to add single blocks to the page content?

site/src/common/blocks/CallToActionListBlock.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { CallToActionListBlockData } from "@src/blocks.generated";
+import { CallToActionBlock } from "@src/common/blocks/CallToActionBlock";
+import styled from "styled-components";
+
+export const CallToActionListBlock = withPreview(
+    ({ data: { blocks } }: PropsWithData<CallToActionListBlockData>) => (
+        <Root>
+            {blocks.map((block) => (
+                <CallToActionBlock data={block.props} key={block.key} />
+            ))}
```
johnnyomair, line 12: Please use the `ListBlock` helper from `@comet/cms-site` here

site/src/common/blocks/TextLinkBlock.tsx
johnnyomair, line 1: I don't think we should delete this (yet), as it maybe useful for other blocks/components, for instance, the footer

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,100 @@
...
+type ButtonVariant = "Contained" | "Outlined" | "Text";
+
+type ButtonProps = {
+    variant?: ButtonVariant;
+    onClick?: () => void;
+    disabled?: boolean;
+    children?: ReactNode;
+} & (HTMLAttributes<HTMLButtonElement> | (HTMLAttributes<HTMLAnchorElement> & Pick<HTMLAnchorElement, "href" | "target">));
+
+export const Button = ({ variant = "Outlined", disabled = false, children, ...htmlAttributes }: ButtonProps) => {
```
johnnyomair, line 13: I believe you'll need ref-forwarding here, otherwise Next will show an error for internal links.

site/src/theme.ts
johnnyomair: MUI somehow supports autocomplete while not having the underscore. Could we do the same?

image

admin/src/common/blocks/CallToActionListBlock.tsx
```diff
@@ -0,0 +1,11 @@
+import { createListBlock } from "@comet/blocks-admin";
+import { CallToActionBlock } from "@src/common/blocks/CallToActionBlock";
+import { FormattedMessage } from "react-intl";
+
+export const CallToActionListBlock = createListBlock({
+    name: "CallToActionList",
+    displayName: <FormattedMessage id="callToActionListBlock.displayName" defaultMessage="Call To Action List" />,
+    block: CallToActionBlock,
+    itemName: <FormattedMessage id="callToActionListBlock.itemName" defaultMessage="Call To Action" />,
+    itemsName: <FormattedMessage id="callToActionListBlock.itemsName" defaultMessage="Call To Action List" />,
```
johnnyomair, line 10: This should probably be `action` and `actions` 🤔

site/src/theme.ts
jamesricky: > MUI somehow supports autocomplete while not having the underscore. Could we do the same?

Just tried it, seems to already work anyway in VSCode and PhpStorm 😅

api/src/documents/pages/blocks/page-content.block.ts
SebiVPS: @thomasdax98 what do you think?

api/src/common/blocks/call-to-action.block.ts
SebiVPS: hmm, i think its clearer when its named what it is. there you know you also have the text component 🤔

site/src/common/blocks/CallToActionListBlock.tsx
SebiVPS: nice 👍  did not know that one yet

site/src/common/blocks/TextLinkBlock.tsx
SebiVPS: ok i added it again, but its not used anywhere atm.

api/src/documents/pages/blocks/page-content.block.ts
SebiVPS: ahh, now i understand. Because i did not added a single block to the admin PageContentBlock, i do not need it here. 👍 
I removed it here.

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,122 @@
+import { forwardRef, HTMLAttributes, ReactNode, RefObject } from "react";
+import styled, { css } from "styled-components";
+
+type ButtonVariant = "Contained" | "Outlined" | "Text";
```
jamesricky, line 4: We always define values like this in `camelCase`, this should probably be the same here. 

```suggestion
type ButtonVariant = "contained" | "outlined" | "text";
```

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+type ButtonProps = {
+    variant?: ButtonVariant;
+    onClick?: () => void;
+    disabled?: boolean;
+    children?: ReactNode;
```
jamesricky, line 10: I think these values should already be defined by the `HTMLAttributes` or by the element type. 
Just like `href` for instance. 

```suggestion
```

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,122 @@
...
+        `}
+    `,
+    Text: css`
+        color: ${({ theme }) => theme.palette.grey["300"]};
+    `,
+};
+
+const commonButtonStyle = css`
+    display: inline-flex;
+    box-sizing: border-box;
```
jamesricky, line 86: `border-box` doesn't do anything, unless there is a `width`/`height` value in addition to the `padding`. 

```suggestion
```

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,122 @@
...
+    padding: ${({ theme }) => `${theme.spacing.S400} ${theme.spacing.S500}`};
+    border-radius: 4px;
+    cursor: pointer;
+    justify-content: center;
+    align-items: center;
+    transition: all 0.2s ease-out;
```
jamesricky, line 92: I would avoid transitioning all properties. 
This could cause unintended animations, e.g. when extending the button, adding responsive styles, like a different `padding` or `font-size`.

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,122 @@
+import { forwardRef, HTMLAttributes, ReactNode, RefObject } from "react";
+import styled, { css } from "styled-components";
+
+type ButtonVariant = "Contained" | "Outlined" | "Text";
+
+type ButtonProps = {
+    variant?: ButtonVariant;
```
jamesricky, line 7: @mariokemetinger @johnnyomair do you think a `fullWidth` prop would make sense, to add `width: 100%` or `display: flex` instead of `inline-flex`? 

This would be useful, when the buttons size should depend on the parent.

site/src/components/common/Button.tsx
SebiVPS: i changed the transition to background-color, is this ok?

site/src/components/common/Button.tsx
SebiVPS: i think we can add it, when the use-case is given.

mariokemetinger (approved):

johnnyomair (approved):

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+    ${({ $variant, $disabled }) =>
+        $disabled &&
+        css`
+            pointer-events: none;
+            ${disabledButtonVariantStyle[$variant]};
+        `}
+`;
+
+const StyledButton = styled.button<{ $variant: ButtonVariant }>`
```
nsams, line 112: Can't you use `as` prop for supporting a and button instead of the two Styled components?

SebiVPS: It was implemented like that before. @johnnyomair  mentioned that the button component already have a disabled prop, which should also be implemented in the css with `:disabled {}` that is why i split them up. 
Do you think of a better solution for that?

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,120 @@
...
+    text-decoration: none;
+    font-family: ${({ theme }) => theme.fontFamily};
+    font-size: 16px;
+    font-weight: 700;
+    line-height: 110%;
+`;
+
+const StyledAnchor = styled.a<{ $variant: ButtonVariant; $disabled: boolean }>`
+    ${commonButtonStyle};
+    ${({ $variant }) => buttonVariantStyle[$variant]};
```
jamesricky, line 102: You could move this to `commonButtonStyles` as well, when doing this:

```ts
const commonButtonStyle = css<{ $variant: ButtonVariant }>`
    // ...
`;
```

site/src/components/common/Button.tsx
jamesricky: I think it's fine to have two styled components for the the individual styles.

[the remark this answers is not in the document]
jamesricky: `ButtonRoot` and `AnchorRoot`? 🤔  

I'm fine with `Styled...` in this case thought.

site/src/components/common/Button.tsx
jamesricky: Since we use `camelCase` for everything else like this in the site, it should be the same here imo, even if it requires additional mapping. 

@johnnyomair FYI: This is one of the reasons I don't like enums in combination with our coding-guidelines, requiring enums to be `PascalCase` 😄

site/src/components/common/Button.tsx
jamesricky: If that's all that should be animates: yes. 
But `border-color` and `color` are also changed on `:hover`, should they not be animated?

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,120 @@
...
+const buttonVariantStyle: Record<ButtonVariant, ReturnType<typeof css>> = {
+    Contained: css`
+        ${({ theme }) => css`
+            background-color: ${theme.palette.primary.main};
+            color: ${theme.palette.primary.contrastText};
+            border: 1px solid ${theme.palette.primary.main};
+
+            &:hover {
+                background-color: ${theme.palette.primary.dark};
+                border: 1px solid ${theme.palette.primary.dark};
```
jamesricky, line 35: You could just change the `border-color` here, as the rest does not change. 
E.g. if the width ever changes in a project, the value only has to be changed in one place then. 

```suggestion
                border-color: ${theme.palette.primary.dark};
```

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,120 @@
...
+    `,
+    Outlined: css`
+        ${({ theme }) => css`
+            background-color: transparent;
+            color: ${theme.palette.primary.main};
+            border: 1px solid ${theme.palette.primary.main};
+
+            &:hover {
+                color: ${theme.palette.primary.dark};
+                border: 1px solid ${theme.palette.primary.dark};
```
jamesricky, line 47:
```suggestion
                border-color: ${theme.palette.primary.dark};
```

site/src/components/common/Button.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const commonButtonStyle = css<{ $variant: ButtonVariant }>`
+    display: inline-flex;
+    padding: ${({ theme }) => `${theme.spacing.S400} ${theme.spacing.S500}`};
+    border-radius: 4px;
+    cursor: pointer;
+    justify-content: center;
+    align-items: center;
+    transition: all 0.2s ease-out;
+    transition-property: background-color, color, border-color;
```
jamesricky, line 91:
```suggestion
    transition: background-color 0.2s ease-out, color 0.2s ease-out, border-color 0.2s ease-out;
```

site/src/components/common/Button.tsx
nsams: why, do you think using `as` is too complicated?

jamesricky (approved):

site/src/components/common/Button.tsx
johnnyomair: The "problem" is that `button` supports `disabled`, whereas `a` doesn't. I'm not sure if this can be achieved with one styled-component, but even if it can, it adds more complexity than simply using two.

johnnyomair (approved):

mariokemetinger (approved):

PR #253, opened 2024-05-24
COM-769: Remove @mui/icons-material dependency from admin

jomunker:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-769


dkarnutsch (approved):

nsams (approved):

johnnyomair (approved):
We may merge it like this, but this will only have an effect once we've removed the peer dependency in the core packages.

jamesricky (approved):

PR #255, opened 2024-06-03
Add page layout style

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-714


nsams:
IMHO a Component should never take care about it's own positioning, it should only position it's (direct) children.

And a component should be possible to be render anywhere.

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -117,17 +118,23 @@ const defaultRenderers: Renderers = {
...
+        const content = (
             <PreviewSkeleton title="RichText" type="rows" hasContent={hasRichTextBlockContent(data)}>
                 {rendered}
             </PreviewSkeleton>
         );
+
+        if (isNested) {
+            return content;
+        }
+        return <BlockRoot>{content}</BlockRoot>;
```
SebiVPS, line 137: @nsams i tried not have this condition in the blocks. Now the BlockRoot is clean, but everywhere we have this case, the condition is needed.

site/src/components/common/BlockRoot.tsx
```diff
@@ -0,0 +1,28 @@
+import { gridTotalColumns } from "@src/theme";
+import { HTMLAttributes, ReactNode } from "react";
+import styled, { css } from "styled-components";
+
+interface BlockRootProps extends HTMLAttributes<HTMLDivElement> {
+    grid?: boolean;
+    children?: ReactNode;
```
jamesricky, line 7: Is `children` not already included in `HTMLAttributes<HTMLDivElement>`?

```suggestion
```

site/src/theme.ts
```diff
@@ -5,6 +5,8 @@ const createBreakpoint = (value: number) => {
     };
 };
 
+export const gridTotalColumns = 24;
+
 export const theme = {
```
jamesricky, line 10: Should this just be inside the theme, if it's already in the theme file?

```suggestion
export const theme = {
    gridTotalColumns: 24,
```

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -117,17 +118,23 @@ const defaultRenderers: Renderers = {
 
 interface RichTextBlockProps extends PropsWithData<RichTextBlockData> {
     renderers?: Renderers;
+    isNested?: boolean;
```
jamesricky, line 121: I think this should be named in a way that makes it more obvious what this does or why it is needed. 
Maybe `disableBlockRoot`, `disableSideSpacing`, `disableContentSpacing` 🤔

site/src/components/common/BlockRoot.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const BlockRoot = ({ grid, children, ...restProps }: BlockRootProps) => (
```
jamesricky, line 10: This component doesn't really do anything. 

Could you just export the styled component directly?
Then you also don't need the props type.

site/src/components/common/BlockRoot.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <Root $grid={grid} {...restProps}>
+        {children}
+    </Root>
+);
+
+const Root = styled.div<{ $grid?: boolean }>`
```
jamesricky, line 16: Would a `$disableSideSpacing` or `$disableSidePadding` prop be useful? 
This could be used instead of conditional rendering in e.g. what is now done with `isNested` in `RichTextBlock`.

site/src/components/common/BlockRoot.tsx
SebiVPS: I think you are right, this does nothing anymore.
I'll try another solution tomorrow to just add the css for the block root div to the theme (max-width, padding, ...)
The two style props for the grid root are not worth to outsource anywhere, i guess.

site/src/theme.ts
SebiVPS: It makes things more complicated because the variable is used outside the theme context. Then you have to manually import the whole theme from this file.

site/src/components/common/BlockRoot.tsx
SebiVPS: @jamesricky new approach commited, also used in #257

SebiVPS: i think about closing this PR and only merge #257  (the styling refactors and grid stylings are not used here) 
only the page layout style seems a bit less for that PR.

PR #256, opened 2024-06-05
Add dynamic paragraph spacing

SebiVPS:
Also add `$disableLastBottomSpacing` for the rich text block to add the possibility to disable the last bottom margin when to block is used within another block.

### DIFF
--- a/site/src/common/blocks/RichTextBlock.tsx
+++ b/site/src/common/blocks/RichTextBlock.tsx
@@ -28,55 +28,55 @@ export const defaultRichTextRenderers: Renderers = {
     blocks: {
         unstyled: (children, { keys }) =>
             children.map((child, index) => (
-                <Text key={keys[index]} gutterBottom>
+                <Text key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "paragraph-standard": (children, { keys }) =>
             children.map((child, index) => (
-                <Text key={keys[index]} gutterBottom>
+                <Text key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "paragraph-small": (children, { keys }) =>
             children.map((child, index) => (
-                <Text variant="p200" key={keys[index]} gutterBottom>
+                <Text variant="p200" key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "header-one": (children, { keys }) =>
             children.map((child, index) => (
-                <Text variant="h600" key={keys[index]} gutterBottom>
+                <Text variant="h600" key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "header-two": (children, { keys }) =>
             children.map((child, index) => (
-                <Text variant="h550" key={keys[index]} gutterBottom>
+                <Text variant="h550" key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "header-three": (children, { keys }) =>
             children.map((child, index) => (
-                <Text variant="h500" key={keys[index]} gutterBottom>
+                <Text variant="h500" key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "header-four": (children, { keys }) =>
             children.map((child, index) => (
-                <Text variant="h450" key={keys[index]} gutterBottom>
+                <Text variant="h450" key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "header-five": (children, { keys }) =>
             children.map((child, index) => (
-                <Text variant="h400" key={keys[index]} gutterBottom>
+                <Text variant="h400" key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
         "header-six": (children, { keys }) =>
             children.map((child, index) => (
-                <Text variant="h350" key={keys[index]} gutterBottom>
+                <Text variant="h350" key={keys[index]} addMarginBottom>
                     {child}
                 </Text>
             )),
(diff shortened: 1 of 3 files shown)


site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -117,21 +117,36 @@ const defaultRenderers: Renderers = {
...
             <PreviewSkeleton title="RichText" type="rows" hasContent={hasRichTextBlockContent(data)}>
-                {rendered}
+                <Wrapper $disableLastGutterBottom={disableLastGutterBottom}>{rendered}</Wrapper>
             </PreviewSkeleton>
         );
     },
     { label: "Rich Text" },
 );
 
+const Wrapper = styled.div<{ $disableLastGutterBottom?: boolean }>`
```
jamesricky, line 136: We generally name the outer-most element of a component `Root`. 

```suggestion
const Root = styled.div<{ $disableLastGutterBottom?: boolean }>`
```

mariokemetinger (dismissed):

site/src/components/common/Typography.tsx
```diff
@@ -151,6 +213,15 @@ const Text = styled.div<TextProps>`
...
     margin-top: 0;
+
+    ${({ theme, gutterBottom }) =>
+        !gutterBottom &&
+        css`
+            margin-bottom: 0;
+
+            ${theme.breakpoints.xs.mediaQuery} {
+                margin-bottom: 0;
+            }
```
johnnyomair, line 225: There are some typography variants that set a bottom margin for breakpoints larger than XS (e.g., P200). Wouldn't the margin then still be shown for those larger breakpoints?

SebiVPS: No, since this media query i active for all greater values, it overrides these, because it is arranged below.

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -116,21 +116,36 @@ export const defaultRichTextRenderers: Renderers = {
 
 interface RichTextBlockProps extends PropsWithData<RichTextBlockData> {
     renderers?: Renderers;
+    disableLastGutterBottom?: boolean;
```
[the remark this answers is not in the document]
SebiVPS: This was requested by @jamesricky. Gutter is the name used by MUI, that is why it is called this way. (technically it is the same like a margin)

The use-case is, when the richTextBlock is used in custom blocks, and the bottom margin is handled by this component itself, one should be able to disable it.

site/src/components/common/Typography.tsx
nsams: It is a css specificity issue?

SebiVPS: What do you mean with issue?
I thought i can simply override the margin-bottom without the media query, but that was not the case, i had to also add an media query that is active below the media queries before. 
I can also add margin-bottom: 0; for all (greater) other breakpoints, but they will just override themselves so i just added the smallest one.
If you call that an issue, than it is one :)

nsams: > What do you mean with issue?

When you have two places in your code that apply the same style (with different values) to the same element there is this specificity algorithm that calculates by various rules which css declarations is used. This is often a bit complicated and needs techniques to work around "issues".

In your case I suspect (pseudo code):
- "normal" styling: `margin-bottom: 10px`
- "normal" styling: `@media(...)  { margin-bottom: 5px }
- "disableLastGutterBottom" styling, in css code after "normal": `margin-bottom: 0`

now :two: wins over :three:, as it has a higher specificity (by the additional @media) Adding a second `@media(...) { margin-bottom:0 }` will override it, as it has the same specificity and is last in css code. By that you solved the "css specificity issue"

site/src/common/blocks/RichTextBlock.tsx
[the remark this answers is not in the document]
nsams: I would call it margin, why use a different word.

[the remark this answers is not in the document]
jamesricky: If we change the name of `disableLastGutterBottom` here, I think we should also change the name of the `gutterBottom` prop in `Typography`. 

I don't really like the name `gutterBottom` from MUI anyway, but:

- a `marginBottom` prop would make it seem, like it's possible to set the number value, instead of enabling the variant-dependant spacing value as a boolean.
- `gutterBottom` just defines that a spacing is implemented, not how it's implemented. Helpful to have a generic name for this, that can also be used e.g. in mails or native apps, where `margin` may not be possible/optimal. 

We could change it to `bottomSpacing`:
- `gutterBottom` -> `bottomSpacing`
- `disableLastGutterBottom` -> `disableLastBottomSpacing`

Or should we just use `margin` anyway, just make it obvious it's a boolean, not a number?
- `gutterBottom` -> `enableMarginBottom`
- `disableLastGutterBottom` -> `disableLastMarginBottom`

jamesricky (dismissed):

site/src/common/blocks/RichTextBlock.tsx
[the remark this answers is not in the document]
johnnyomair: How about `addMarginBottom` and `disableLastMarginBottom`?

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -116,21 +116,36 @@ export const defaultRichTextRenderers: Renderers = {
 
 interface RichTextBlockProps extends PropsWithData<RichTextBlockData> {
     renderers?: Renderers;
+    disableLastMarginBottom?: boolean;
```
thomasdax98, line 119:
```suggestion
    disableLastBottomSpacing?: boolean;
```

site/src/common/blocks/RichTextBlock.tsx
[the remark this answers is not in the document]
nsams: spacing, another word for the same thing :) But I understand your argument.

@johnnyomair's names are good!

jamesricky (approved):

mariokemetinger (approved):

johnnyomair (approved):

PR #257, opened 2024-06-07
Columns block, page grid layout

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-834, COM-714

<details>
Screenshot 2024-06-07 at 11 25 31

Screenshot 2024-06-07 at 11 25 41

Screenshot 2024-06-07 at 11 26 23

Screenshot 2024-06-11 at 15 28 03

</details>


site/src/layout/GlobalStyles.ts
```diff
@@ -0,0 +1,80 @@
...
+export const GlobalStyle = createGlobalStyle`
+    /* Fix a problem with Flexbox to avoid overflows: https://defensivecss.dev/tip/flexbox-min-content-size/ */
+    * {
+        min-width: 0;
+    }
+    
+    html {
+        --defaultMaxWidth: 1600px;
+        --vwDiff: calc(100vw - 100%);
+        ${gridColumnsCSSVars()};
```
nsams, line 14: could you explain those variables? And why is defaultMaxWidth a css variable?

site/src/documents/pages/blocks/PageContentBlock.tsx
```diff
@@ -13,13 +14,14 @@ import { TextImageBlock } from "@src/common/blocks/TextImageBlock";
 const supportedBlocks: SupportedBlocks = {
     anchor: (props) => <AnchorBlock data={props} />,
     space: (props) => <SpaceBlock data={props} />,
-    richtext: (props) => <RichTextBlock data={props} />,
+    richtext: (props) => <RichTextBlock data={props} addPageGridLayoutStyle />,
```
nsams, line 17: why does a richtext block need a PageGridLyout nd other blocks not?

admin/src/common/blocks/ColumnsContentBlock.tsx
```diff
@@ -0,0 +1,11 @@
+import { createBlocksBlock } from "@comet/blocks-admin";
+import { RichTextBlock } from "@src/common/blocks/RichTextBlock";
+import { SpaceBlock } from "@src/common/blocks/SpaceBlock";
+
+export const ColumnsContentBlock = createBlocksBlock({
+    name: "ColumnsContent",
+    supportedBlocks: {
+        richtext: RichTextBlock,
+        space: SpaceBlock,
```
nsams, line 9: shouldn't there be almost all PageContent blocks?

site/src/layout/GlobalStyles.ts
SebiVPS: @mariokemetinger  can explain you these variables in detail. (https://codepen.io/MarioKemetinger/pen/xxBJWrm) (also see COM-714)

You are right, the defaultMaxWidth must not be a css variable, it is implemented in another project this way, but it do not see why. I'll change that 👍

site/src/documents/pages/blocks/PageContentBlock.tsx
SebiVPS: The other block are not refactored yet. Just added the merged HeadingBlock right now.

admin/src/common/blocks/ColumnsContentBlock.tsx
SebiVPS: The other block are not refactored yet. Just added the merged HeadingBlock right now.

site/src/layout/GlobalStyles.ts
nsams: yes, please explain the ideas behind this

site/src/documents/pages/blocks/PageContentBlock.tsx
```diff
@@ -13,13 +14,14 @@ import { TextImageBlock } from "@src/common/blocks/TextImageBlock";
 const supportedBlocks: SupportedBlocks = {
     anchor: (props) => <AnchorBlock data={props} />,
     space: (props) => <SpaceBlock data={props} />,
-    richtext: (props) => <RichTextBlock data={props} />,
-    heading: (props) => <HeadingBlock data={props} />,
+    richtext: (props) => <RichTextBlock data={props} addPageGridLayoutStyle />,
+    heading: (props) => <HeadingBlock data={props} addPageGridLayoutStyle />,
```
SebiVPS, line 18: @nsams @mariokemetinger @jamesricky @johnnyomair 
An alternative i thought about, was to add a wrapping div with the pageGridLayoutStyle here, instead of adding it in the blocks inside. 
Technically it would be cleaner to add it here. What do you think?

site/src/layout/GlobalStyles.ts
SebiVPS: Just discussed this topic with @mariokemetinger, we are going to remove these variables for now from the starter. 👍

nsams:
Isn't this still moving the positioning of blocks into the blocks? (as in https://github.com/vivid-planet/comet-starter/pull/255#issuecomment-2152076950)

SebiVPS:
> Isn't this still moving the positioning of blocks into the blocks? (as in [#255 (comment)](https://github.com/vivid-planet/comet-starter/pull/255#issuecomment-2152076950))

It is, the only other solution would be something like that:
https://github.com/vivid-planet/comet-starter/pull/257#discussion_r1632775336

I talked to Mario earlier today, he thinks it is ok. But we can discuss this point. What do you think?

site/src/documents/pages/blocks/PageContentBlock.tsx
nsams: yes! Or a wrapper around PageContentBlock.

nsams: > but not practicable because there are use-cases (eg. background-color) for blocks which need to be set for the full width

please try to include an block in this PR where this use-case is actually used

site/src/util/PageGridLayoutStyle.ts
```diff
@@ -0,0 +1,14 @@
+import { gridTotalColumns } from "@src/theme";
+import { css } from "styled-components";
+
+export const pageGridLayoutStyle = css`
```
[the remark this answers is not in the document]
SebiVPS: Thats all whats left.

site/src/components/common/ConditionalPageGridLayout.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { ReactNode } from "react";
+import styled from "styled-components";
+
+interface ConditionalPageGridLayoutProps {
+    pageGridLayout?: boolean;
+    gridColumn?: string;
+    children?: ReactNode;
+}
+
+export const ConditionalPageGridLayout = ({ pageGridLayout, gridColumn, children }: ConditionalPageGridLayoutProps) => {
```
nsams, line 11: One component that does based on props as 3 different things: not easy to use/understand.

I would prefer something simple: for example a GridRoot component and a GridColumn component.

SebiVPS: Yes you are right, something like that would optimize the usage and outsource the conditional rendering, but would be too difficult to understand/use.
I refactored once again to a simple GridRoot, which can we put into the PageContentBlock in comet v7 (then we do not need the conditional render in the blocks).
I think thats the best solution i can think of atm. Maybe we can come up with something magic i the todays discussion.

site/src/components/common/GridRoot.ts
```diff
@@ -0,0 +1,11 @@
+import { gridTotalColumns } from "@src/theme";
+import styled from "styled-components";
+
+export const GridRoot = styled.div`
+    max-width: ${({ theme }) => theme.defaultMaxWidth};
+    margin: 0 auto;
+    box-sizing: border-box;
```
jamesricky, line 7:
```suggestion
```

site/src/common/blocks/CallToActionListBlock.tsx
```diff
@@ -23,3 +34,7 @@ const Root = styled.div`
         gap: ${({ theme }) => theme.spacing.S400};
     }
 `;
+
+const GridColumn = styled.div`
+    grid-column: 3 / 23;
```
jamesricky, line 39: I would prefer `3 / -3`. Makes it easier to see that it's centered / indented symmetrically. 
Also this would still work if `grid-template-columns` is changed to some other value. 

_(same for the other blocks)_

 ```suggestion
     grid-column: 3 / -3;
```

File: site/src/common/blocks/CallToActionListBlock.tsx [hunk 9]
```diff
@@ -23,3 +34,7 @@ const Root = styled.div`
         gap: ${({ theme }) => theme.spacing.S400};
     }
 `;
+
+const GridColumn = styled.div`
```
jamesricky, line 38: Since this is explicitly used only for the `PageContent` maybe this should be reflected in the name, rather than naming it after the used css property...

Maybe: `PageLayoutContent`, `PageIndentation`, `PageContentIndentation` 🤔 

_(same for the other blocks)_

jamesricky (approved):

site/src/common/blocks/CallToActionListBlock.tsx
```diff
@@ -2,17 +2,28 @@
 import { ListBlock, PropsWithData, withPreview } from "@comet/cms-site";
 import { CallToActionListBlockData } from "@src/blocks.generated";
 import { CallToActionBlock } from "@src/common/blocks/CallToActionBlock";
+import { PageLayout } from "@src/components/common/PageLayout";
```
johnnyomair, line 5: Please move this file to either `src/common/PageLayout` or `src/layout/PageLayout`. In Starter we group files by feature (products, pages, common), not by type (hooks, components)

site/src/common/blocks/ColumnsBlock.tsx
```diff
@@ -0,0 +1,50 @@
...
+    "12-6": "17 / 23",
+    "6-12": "11 / 23",
+};
+
+export const ColumnsBlock = withPreview(
+    ({ data: { columns, layout } }: PropsWithData<ColumnsBlockData>) => {
+        return (
+            <PageLayout grid>
+                {columns.map((column, index) => {
+                    const gridColumn = index === 0 ? column1Grid[layout] : column2Grid[layout];
```
johnnyomair, line 27: This doesn't seem right to me... Could you please further explain? What about adding two columns in a one-column-layout? In our projects, we usually showed them underneath.

site/src/common/blocks/ColumnsContentBlock.tsx
```diff
@@ -0,0 +1,22 @@
...
+    space: (props) => <SpaceBlock data={props} />,
+    heading: (props) => <HeadingBlock data={props} />,
+    callToActionList: (props) => <CallToActionListBlock data={props} />,
+};
+
+export const ColumnsContentBlock = withPreview(
+    ({ data }: PropsWithData<ColumnsContentBlockData>) => {
+        return <BlocksBlock data={data} supportedBlocks={supportedBlocks} />;
+    },
+    { label: "Column" },
```
johnnyomair, line 21:
```suggestion
    { label: "Columns" },
```

site/src/components/common/PageLayout.tsx
```diff
@@ -0,0 +1,25 @@
+import { gridTotalColumns } from "@src/theme";
+import { HTMLAttributes, PropsWithChildren } from "react";
+import styled, { css } from "styled-components";
+
+interface PageLayoutProps extends PropsWithChildren<HTMLAttributes<HTMLDivElement>> {
```
johnnyomair, line 5: I believe PropsWithChildren is redundant here

site/src/components/common/PageLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+    grid?: boolean;
+}
+
+export const PageLayout = ({ grid, children, ...restProps }: PageLayoutProps) => (
+    <Root $grid={grid} {...restProps}>
+        {children}
+    </Root>
```
johnnyomair, line 12:
```suggestion
export const PageLayout = ({ grid, ...restProps }: PageLayoutProps) => (
    <Root $grid={grid} {...restProps} />
```

johnnyomair (changes requested):

johnnyomair (commented):
Please resolve conflicts

site/src/common/blocks/ColumnsBlock.tsx
SebiVPS: I tried to simplify the usage of the columnGrid map.

> What about adding two columns in a one-column-layout? In our projects, we usually showed them underneath.

As far as i know there are no plans in wrapping a column layout in another column layout. 
If that is really needed one has to refactor this component completely,  an additional logic is also needed to get the information whats the parent columns layout?

SebiVPS: I think, found the perfect solution: using the modulo operator it can be ensured that the additional columns block are rendered underneath:
Screenshot 2024-06-18 at 12 04 25

SebiVPS: `$column={columnGrid[index % Object.keys(columnGrid).length][layout]}`

that line is indeed a bit hard to read 🤔  maybe you have a suggestion to improve the readability?

SebiVPS: Ok, i was wrong, modulo is not the right choice, it does not work for all cases. I'll think about it after lunch 🤔

SebiVPS: With an additional map, where i have access to the column count it is possible with the modulo operator 👍 
What do you think about this solution? Is there a way to improve this a bit more?

johnnyomair: Tried it myself in #262. IMO this is a cleaner solution, what do you think?

SebiVPS: Yes 👍  that looks a lot cleaner.

johnnyomair (dismissed):
LGTM.

johnnyomair (approved):

jamesricky (approved):

PR #258, opened 2024-06-11
Use site-configs

fraxachun:
https://vivid-planet.atlassian.net/browse/COM-855


site-configs.local.ts
```diff
@@ -0,0 +1,49 @@
+import { SiteConfig } from "./site-configs.types";
+
+// Fallback, if directory ./site-configs exists, this file will not be used and should be deleted
```
nsams, line 3: I don't understand why you made this fallback

package.json
```diff
@@ -23,6 +24,7 @@
         "browser:jaeger": "dotenv -c secrets -- sh -c 'open-cli http://localhost:$JAEGER_UI_PORT'"
     },
     "devDependencies": {
+        "@comet/cli": "7.0.0-canary-20240611065117",
```
nsams, line 27: does this not work?
```suggestion
        "@comet/cli": "workspace:*",
```

admin/src/App.tsx
```diff
@@ -55,10 +54,19 @@ export function App() {
...
             <ApolloProvider client={apolloClient}>
                 <BuildInformationProvider value={{ date: config.buildDate, number: config.buildNumber, commitHash: config.commitSha }}>
-                    <SitesConfigProvider
+                    <SitesConfigProvider<SitesConfig>
                         value={{
                             configs: config.sitesConfig,
-                            resolveSiteConfigForScope: (configs, scope: ContentScope) => configs[scope.domain],
+                            resolveSiteConfigForScope: (configs, scope: ContentScope) => {
+                                const siteConfig = configs.find((config) => config.contentScope.domain === scope.domain);
+                                if (!siteConfig) throw new Error(`siteConfig not found for country ${scope.domain}`);
```
nsams, line 62:
```suggestion
                                if (!siteConfig) throw new Error(`siteConfig not found for domain ${scope.domain}`);
```

admin/src/App.tsx
```diff
@@ -55,10 +54,19 @@ export function App() {
         <ConfigProvider config={config}>
             <ApolloProvider client={apolloClient}>
                 <BuildInformationProvider value={{ date: config.buildDate, number: config.buildNumber, commitHash: config.commitSha }}>
-                    <SitesConfigProvider
+                    <SitesConfigProvider<SitesConfig>
```
nsams, line 57: couldn't you infer that from passed `configs`?

admin/src/common/ContentScopeProvider.tsx
```diff
@@ -10,16 +10,8 @@ import {
...
 } from "@comet/cms-admin";
-import { SitesConfig } from "@src/config";
-
-type Domain = "main" | "secondary" | string;
-type Language = "en" | string;
-export interface ContentScope {
-    domain: Domain;
-    language: Language;
-}
+import { ContentScope } from "@src/config";
```
nsams, line 14: :+1:

package.json
nsams: oops :)

admin/src/App.tsx
nsams: This is not what I mean, I was thinking if that could work without explicit generic parameter:
```
<SitesConfigProvider
                        value={{
                            configs: config.sitesConfig,
                            resolveSiteConfigForScope: (configs, scope: ContentScope) => {
```

see here a (somehow) working example:
https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAZRlApgZwMIODAcwAUAnOANxgBN0SAeXMfAzAPgAoAHMzzALkQBvROQCGAGxDoBwiHkL9EjZpgA0iEljjjy6VBmWEAYnBLI5naYnZymCgYZYBKRAF5WiUWACeAX18uggCwAFChvqF0+liOpBTUtCISUq6CgrYqMojAcHACAIwATADMvuqamNq60Y4mZhZWNvLObh7BIYidGuhQICRIGQoAdDlwANzh-qyRAPTROM1xlDQkrEA

api/src/config/config.ts
```diff
@@ -2,6 +2,10 @@ import cometConfig from "@src/comet-config.json";
 import { plainToClass } from "class-transformer";
 import { validateSync } from "class-validator";
 
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+// eslint-disable-next-line
+import type { PrivateSiteConfig } from "../../../site-configs.types";
```
nsams, line 8: does `@src/../../` not work here?

api/src/config/environment-variables.ts
```diff
@@ -115,4 +115,7 @@ export class EnvironmentVariables {
     @IsString()
     @ValidateIf(() => process.env.NODE_ENV === "production")
     CDN_ORIGIN_CHECK_SECRET: string;
+
+    @IsString()
```
dkarnutsch, line 119: Can we use IsJson here to detect errors?

admin/package.json
```diff
@@ -4,8 +4,8 @@
     "scripts": {
         "postinstall": "cd server && npm install",
         "serve": "cd server && npm run serve",
-        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -- vite",
-        "build": "npm run intl:compile && run-p gql:types generate-block-types && NODE_ENV=production dotenv -c secrets -- vite build",
+        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -e .env.site-configs -- vite",
+        "build": "npm run intl:compile && run-p gql:types generate-block-types && NODE_ENV=production dotenv -c secrets -e .env.site-configs -- vite build",
```
johnnyomair, line 8:
```suggestion
        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -e site-configs -- vite",
        "build": "npm run intl:compile && run-p gql:types generate-block-types && NODE_ENV=production dotenv -c secrets -e site-configs -- vite build",
```

Shouldn't this be just `site-configs` (as for `secrets`)?

admin/src/common/ContentScopeProvider.tsx
```diff
@@ -47,20 +39,17 @@ export function useContentScopeConfig(p: ContentScopeConfigProps): void {
 }
 
 export const ContentScopeProvider: React.FC<Pick<ContentScopeProviderProps, "children">> = ({ children }) => {
-    const sitesConfig = useSitesConfig<SitesConfig>();
```
johnnyomair, line 50: Could it still be that a user might have access to a content scope that isn't part of the sites config yet?

api/nest-cli.json
```diff
@@ -1,4 +1,5 @@
 {
     "collection": "@nestjs/schematics",
-    "sourceRoot": "src"
+    "sourceRoot": "src",
+    "entryFile": "api/src/main"
```
johnnyomair, line 4: Wouldn't it be better to symlink site-configs.types?

api/package.json
```diff
@@ -5,8 +5,8 @@
         "api-generator": "rimraf 'src/*/generated' && comet-api-generator generate",
         "prebuild": "rimraf dist",
         "build": "nest build",
-        "console": "dotenv -c secrets -- ts-node --transpile-only -r tsconfig-paths/register src/console.ts",
-        "console:prod": "node dist/console.js",
+        "console": "dotenv -c secrets -e .env.site-configs -- ts-node --transpile-only -r tsconfig-paths/register src/console.ts",
+        "console:prod": "node dist/src/console.js",
```
johnnyomair, line 9: dist/src/ is strange 😁

package-lock.json
```diff
@@ -1,11 +1,12 @@
 {
     "name": "starter",
-    "lockfileVersion": 3,
+    "lockfileVersion": 2,
```
johnnyomair, line 3: I believe you used the wrong NPM version when installing. This should stay at 3

site-configs.ts
johnnyomair, line 1: Could we move some parts of this file to the library?

site-configs.ts
```diff
@@ -0,0 +1,37 @@
...
+    } & {
+        [key in Environment]?: string;
+    };
+};
+
+// Called by `npx @comet/cli inject-site-configs`
+const getSiteConfigs = async (env: Environment): Promise<SiteConfig[]> => {
+    const path = `${__dirname}/site-configs`;
+
+    const files = fs.readdirSync(path).filter((file) => !file.startsWith("_"));
```
johnnyomair, line 19: Nit: You could use the promises fs API here

package.json
dkarnutsch, line 27: shouldn't this target next if we use comet v7 here?

admin/src/common/ContentScopeProvider.tsx
johnnyomair: Could this be a problem?

site-configs.ts
nsams: :+1: with that we have flexibility

admin/package.json
nsams: yes a bit confusing to mix -c and -e, maybe changing both to -e would be better

package.json
```diff
@@ -2,7 +2,8 @@
     "name": "starter",
     "private": true,
     "scripts": {
-        "dev": "dotenv -c secrets -- dev-pm start",
+        "create-site-configs-env": "npx @comet/cli inject-site-configs -i .env.site-configs.tpl -o .env.site-configs",
+        "dev": "npm run create-site-configs-env && dotenv -c secrets -- dev-pm start",
```
nsams, line 6: I personally never use this dev script. Would it be an alternative to execute create-site-configs-env in install.sh?

nsams (approved):

admin/src/common/ContentScopeProvider.tsx
johnnyomair: I'm thinking about the case where a user has access to a scope in the API that doesn't exist as site config in the Admin. Previously, we did a comparison between `user.allowedContentScopes` and `sitesConfig.configs` to only show the scope that also have a site config. Now, we only use `user.allowedContentScopes`, and there may be a scope that doesn't have a matching site config. But it's quite possibly that this isn't a real problem.

admin/package.json
```diff
@@ -4,8 +4,8 @@
     "scripts": {
         "postinstall": "cd server && npm install",
         "serve": "cd server && npm run serve",
-        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -- vite",
-        "build": "npm run intl:compile && run-p gql:types generate-block-types && NODE_ENV=production dotenv -c secrets -- vite build",
+        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -e .env.site-configs -- vite",
```
johnnyomair, line 7: If we don't use .env.secrets in the Admin, we should probably also remove the symlink. We previously used this for "semi-secret" secrets such as a Google Maps API key that's injected into the file.

package.json
```diff
@@ -17,12 +18,13 @@
...
-        "browser:admin": "dotenv -c secrets -- sh -c 'open-cli $ADMIN_URL'",
-        "browser:site": "dotenv -c secrets -- sh -c 'open-cli $SITE_URL'",
-        "browser:jaeger": "dotenv -c secrets -- sh -c 'open-cli http://localhost:$JAEGER_UI_PORT'"
+        "browser:api": "dotenv -- sh -c 'open-cli $API_URL/graphql'",
+        "browser:admin": "dotenv -- sh -c 'open-cli $ADMIN_URL'",
+        "browser:site": "dotenv -- sh -c 'open-cli $SITE_URL'",
+        "browser:jaeger": "dotenv -- sh -c 'open-cli http://localhost:$JAEGER_UI_PORT'"
     },
     "devDependencies": {
+        "@comet/cli": "7.0.0-canary-20240611065117",
```
johnnyomair, line 27: You could use beta.2 now

johnnyomair (commented):
Please fix the conflicts introduced by the App Router merge.

johnnyomair (approved):

johnnyomair:
@dkarnutsch @thomasdax98 please review

site-configs.ts
```diff
@@ -0,0 +1,36 @@
+import fs from "fs/promises";
+import { SiteConfig } from "./site-configs.d";
+
+// Types for files in site-configs/
+type Environment = "local" | "dev" | "test" | "staging" | "prod";
+export type Config = Omit<SiteConfig, "domains"> & {
+    preLoginPassword?: string;
```
thomasdax98, line 7: Here it's `preLoginPassword`. In `site-configs.d.ts` it's `preloginPassword` (lowercase l). 

I guess you don't need it here at all since the `SiteConfig` type is extended

site-configs.ts
```diff
@@ -0,0 +1,36 @@
...
+    domains: {
+        preliminary?: string;
+    } & {
+        [key in Environment]?: string;
+    };
+};
```
thomasdax98, line 13: Why do we need this split between site.configs.ts and site-configs.d.ts at all? Can't we define the types in one place?

admin/package.json
```diff
@@ -4,8 +4,8 @@
     "scripts": {
         "postinstall": "cd server && npm install",
         "serve": "cd server && npm run serve",
-        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -c secrets -- vite",
-        "build": "npm run intl:compile && run-p gql:types generate-block-types && NODE_ENV=production dotenv -c secrets -- vite build",
+        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -e .env.secrets -e .env.site-configs -- vite",
+        "build": "npm run intl:compile && run-p gql:types generate-block-types && NODE_ENV=production vite build",
```
thomasdax98, line 8: Isn't the site config needed for the build? Or am I overlooking something?

nsams: build must not use site config, as it depends on the environment

dkarnutsch (approved):

thomasdax98 (approved):

johnnyomair (approved):

PR #259, opened 2024-06-12
Add styling for inline links in rich text block

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-781

<details>
Screenshot 2024-06-13 at 16 37 07

</details>


site/src/components/common/TextLink.tsx
```diff
@@ -0,0 +1,32 @@
+import { HTMLAttributes, ReactNode } from "react";
+import styled, { css } from "styled-components";
+
+type TextLinkProps = {
+    disabled?: boolean;
+    children?: ReactNode;
+} & HTMLAttributes<HTMLAnchorElement>;
+
+export const TextLink = ({ disabled = false, children, ...props }: TextLinkProps) => (
```
nsams, line 9: This is basically a styled a, I think TextLink is not the best name.

And is used for something else than RichtText? Maybe move it there?

site/src/components/common/TextLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+    text-decoration: none;
+
+    ${({ $disabled }) =>
+        $disabled
+            ? css`
+                  pointer-events: none;
+                  color: ${({ theme }) => theme.palette.grey["300"]};
+              `
+            : css`
+                  color: ${({ theme }) => theme.palette.primary.main};
```
nsams, line 25: if you set the color it won't inherit anymore from it's parent. The link can be used in a heading or different p that potentially might have different colors.

SebiVPS: According to the design the link should always have this color.  When there is really a case, where the color should be different, one can adapt this?

site/src/components/common/TextLink.tsx
SebiVPS: Yes we can move it to the RichText. The ticket told to make a component, but i also think its better to put it in the RichText and when its needed somewhere else, one can make a reusable component.
@jamesricky @mariokemetinger  what do you think?

johnnyomair: I believe @jamesricky was strongly in favor of having a standalone component...

site/src/components/common/TextLink.tsx
```diff
@@ -0,0 +1,32 @@
+import { HTMLAttributes, ReactNode } from "react";
+import styled, { css } from "styled-components";
+
+type TextLinkProps = {
+    disabled?: boolean;
+    children?: ReactNode;
```
johnnyomair, line 6: Nit: You can use `React.PropsWithChildren` for this

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -108,7 +110,7 @@ export const defaultRichTextRenderers: Renderers = {
         // key is the entity key value from raw
         LINK: (children, data: LinkBlockData, { key }) => (
             <LinkBlock key={key} data={data}>
-                <a>{children}</a>
+                <TextLink disabled={!isValidLink(data)}>{children}</TextLink>
```
johnnyomair, line 113: Instead of rendering a disabled link, shouldn't we just render a span?

site/src/components/common/TextLink.tsx
nsams: > the design

This is the starter that will be used for all future designs

Imagine a standard paragraph with black font color and a quote pagagraph with gray font color, both used in a RichTextBlock.

SebiVPS: The starter design will be used for all future designs (according to UX).
I can not inherit the color and set the color at the same time.

This is the design template i got:
Screenshot 2024-06-13 at 15 52 19

If we do not want to set a fixed color for a link, we need to talk to UX.
I think it would also be quite tricky to set another color for another paragraph style (the link does not know if he is in a quote paragraph or in an normal one, how would you achieve that?)

johnnyomair: This is a paragraph with [a link](https://www.google.com)

> This is a paragraph with [a link](https://www.google.com)

GitHub has no problems with blue links everywhere 😉

nsams: alright, now I understand. Sorry for this.

jamesricky (approved):

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -106,11 +107,16 @@ export const defaultRichTextRenderers: Renderers = {
...
-        ),
+        LINK: (children, data: LinkBlockData, { key }) => {
+            const isValid = isValidLink(data);
+            return (
+                <LinkBlock key={key} data={data}>
+                    <InlineLink as={isValid ? "a" : "span"} $isValid={isValid}>
+                        {children}
+                    </InlineLink>
+                </LinkBlock>
+            );
```
johnnyomair, line 118: As discussed, please render the span without LinkBlock

johnnyomair (approved):

PR #263, opened 2024-06-20
Add style lint to site

SebiVPS:


site/.stylelintrc.json
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "stylelint-config-standard",
+    "customSyntax": "postcss-styled-syntax",
+    "rules": {
+        "no-invalid-double-slash-comments": null,
+        "block-no-empty": true,
+        "declaration-empty-line-before": null,
+        "declaration-block-no-redundant-longhand-properties": null
+    }
```
johnnyomair, line 9: Could you please explain for each rule why you're disabling it?

johnnyomair (commented):
I think it's a good idea to add this!

site/src/pages/_app.page.tsx
```diff
@@ -77,7 +78,8 @@ const GlobalStyle = createGlobalStyle`
     }
 
     /* Create a root stacking context: https://www.joshwcomeau.com/css/custom-css-reset/#eight-root-stacking-context-9 */
-    #root,
+    /* stylelint-disable selector-id-pattern */
+    #root,
```
jamesricky, line 82: I don't think we should prevent ID selectors. 
Especially, if we need this one by default in every project.

nsams: It's a good idea to never use id selectors, and it should be linted. Exceptions apply, and for those the rule is disabled.

nsams:
Have you considered a eslint based plugin?

SebiVPS:
> Have you considered a eslint based plugin?

Do you have one in your mind? 
@martinrohrmoser did the research, maybe you considered a eslint based plugin?

nsams:
> Do you have one in your mind?

no, I'm just the one that questions everything :)

mariokemetinger (approved):

martinrohrmoser (approved):

dkarnutsch (approved):

johnnyomair:
> Have you considered a eslint based plugin?

I did some research on how to lint css with eslint, but it seems that this is only possible for the most basic things (e.g. formatting). I also haven't found a plugin that incorporates stylelint into eslint. It seems to me that using stylelint standalone is the wat to go.

johnnyomair (approved):

johnnyomair:
I'd like to wait on @jamesricky's approval here

thomasdax98 (approved):

jamesricky (approved):
I think we should add this to admin as well for consistency

johnnyomair:
> I think we should add this to admin as well for consistency. @johnnyomair @thomasdax98 what do you think?

Do we have that much custom CSS in the admin? But yes, we could do that in a follow-up PR.

PR #260, opened 2024-06-14
Add accordion block

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-843, COM-844
-   [x] Provide screenshots/screencasts if the change contains visual changes

<details>
    <summary>Screenshots/screencasts</summary>

Expand/Collapse animation

https://github.com/vivid-planet/comet-starter/assets/42858722/edf89971-3f81-446a-b489-628d889182b2

Accessibility:

https://github.com/vivid-planet/comet-starter/assets/42858722/e713d64a-7942-4242-8fe2-54463d586fcf

In columns:
Screenshot 2024-06-28 at 09 57 25

</details>

PR #266, opened 2024-06-20
Add columns block

SebiVPS:
reopened #257


johnnyomair (dismissed):

site/src/common/blocks/ColumnsBlock.tsx
```diff
@@ -0,0 +1,80 @@
+import { PropsWithData, withPreview } from "@comet/cms-site";
+import { ColumnsBlockData } from "@src/blocks.generated";
+import { ColumnsContentBlock } from "@src/common/blocks/ColumnsContentBlock";
+import { PageLayout } from "@src/layout/PageLayout";
+import styled, { css } from "styled-components";
+
+export const ColumnsBlock = withPreview(
```
nsams, line 7: That should be named PageContentColumnsBlock (becasue it is in common/blocks folder)

site/src/common/blocks/ColumnsContentBlock.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const supportedBlocks: SupportedBlocks = {
+    anchor: (props) => <AnchorBlock data={props} />,
+    richtext: (props) => <RichTextBlock data={props} />,
+    space: (props) => <SpaceBlock data={props} />,
+    heading: (props) => <HeadingBlock data={props} />,
+    callToActionList: (props) => <CallToActionListBlock data={props} />,
+};
+
+export const ColumnsContentBlock = withPreview(
```
nsams, line 17: Why do you put that into a separate file? That might confuse someone of which one to use

site/src/documents/pages/blocks/PageContentBlock.tsx
```diff
@@ -2,24 +2,26 @@
...
     anchor: (props) => <AnchorBlock data={props} />,
     space: (props) => <SpaceBlock data={props} />,
-    richtext: (props) => <RichTextBlock data={props} />,
-    heading: (props) => <HeadingBlock data={props} />,
+    richtext: (props) => <PageContentRichTextBlock data={props} />,
+    heading: (props) => <PageContentHeadingBlock data={props} />,
     image: (props) => <DamImageBlock data={props} />,
     textImage: (props) => <TextImageBlock data={props} />,
     damVideo: (props) => <DamVideoBlock data={props} />,
     youTubeVideo: (props) => <YouTubeVideoBlock data={props} />,
```
nsams, line 22: The other blocks also need a PageContent variant, or?

site/src/theme.ts
```diff
@@ -5,6 +5,8 @@ const createBreakpoint = (value: number) => {
     };
 };
 
+export const gridTotalColumns = 24;
```
nsams, line 8: that might be named pageLayoutGridTotalColumns? And have you considered putting it into the theme?

site/src/theme.ts
```diff
@@ -31,6 +33,7 @@ export const theme = {
         },
     },
     fontFamily: "Arial, sans-serif",
+    defaultMaxWidth: "1920px",
```
nsams, line 36: this should also be somehow named with pageLayout, maybe wrapped in an pageLayout object

site/src/layout/GlobalStyles.ts
johnnyomair, line 1: Please rename to GlobalStyle (as the export)

site/src/theme.ts
SebiVPS: I renamed the variable to `pageLayoutGridTotalColumns`
I considered putting it into the theme but decided against it because i think this should be a project wide setting and should not be changing in different themes (the grid-column values in all associated blocks would also need adaptions if this value changes).
According to that, this file is not really the right place, maybe we put it in PageLayout.tsx directly?

site/src/documents/pages/blocks/PageContentBlock.tsx
SebiVPS: the other blocks will be deleted or replaced

site/src/common/blocks/ColumnsBlock.tsx
SebiVPS: I do not think we should rename the file, that would be confusing, a ColumnsBlock is easier to find.

site/src/common/blocks/ColumnsContentBlock.tsx
SebiVPS: I also did the same in #260 , @johnnyomair @mariokemetinger @jamesricky what do you think, what approach do 
you prefer?

site/src/theme.ts
johnnyomair: > According to that, this file is not really the right place, maybe we put it in PageLayout.tsx directly?

Yes. And only export it once we need the export.

> I considered putting it into the theme but decided against it because i think this should be a project wide setting and should not be changing in different themes (the grid-column values in all associated blocks would also need adaptions if this value changes).

The same argument could also be made for the max-width value, right?

SebiVPS: > > I considered putting it into the theme but decided against it because i think this should be a project wide setting and should not be changing in different themes (the grid-column values in all associated blocks would also need adaptions if this value changes).
> 
> The same argument could also be made for the max-width value, right?

The max-width has no direct dependency to the grid. There is no problem having different max-widths in different themes, that would have no impact to the grid layout and the associated blocks. The total number of the grid columns on the other hand would impact the grid columns values.

If you think it would be better to add it to the page layout as well, and if the case is really there, it can be added to the theme, then we can do it that way as well.

jamesricky (approved):

site/src/common/blocks/ColumnsContentBlock.tsx
johnnyomair: I'd prefer keeping "private" components in the same file as the owning component, but I'm okay with multiple files.

site/src/common/blocks/ColumnsBlock.tsx
johnnyomair: I would only add the distinction between normal and PageContent* blocks if it's necessary (= the block can be used in other places than the page content). For page-content only block I wouldn't add the prefix, but move the blocks to `src/documents/pages/blocks`.

site/src/theme.ts
SebiVPS: also moved the variable to PageLayout.tsx

jamesricky (dismissed):

site/src/common/blocks/ColumnsBlock.tsx
SebiVPS: i moved the file to `src/documents/pages/blocks` 
the only thing that distracts me a bit, is that in the api we do not have this distinction and the block is in `src/common/blocks`

site/src/common/blocks/ColumnsContentBlock.tsx
SebiVPS: ok, i moved the "private" components into the same files, makes sense, when they are not used anywhere else. Just for the sake of clarity additional files in the root block overview list are distracting there.

nsams: > There is no best-practice approach

Please don't look for one best-practice solution: When a component (or file) gets too large and confusing, try to split it up in multiple files. (examples: move parts into a sub component, extract a hook into it's own file, extract styled-components or graphql-queries into their own file, ...) Do **not** do all this too early.

site/src/theme.ts
nsams: > [...] and should not be changing in different themes

what is the plan with theme anyway? what use cases should it solve?

site/src/common/blocks/ColumnsBlock.tsx
johnnyomair: We do! https://github.com/vivid-planet/comet-starter/tree/main/api/src/documents/pages/blocks

site/src/documents/pages/blocks/PageContentBlock.tsx
SebiVPS: todo is added 👍

site/src/common/blocks/ColumnsBlock.tsx
SebiVPS: moved 🚀

nsams (approved):

jamesricky (approved):

mariokemetinger (approved):

johnnyomair (approved):

PR #268, opened 2024-06-24
Add media block

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-772
-   [x] Provide screenshots/screencasts if the change contains visual changes

<details>
    <summary>Screenshots/screencasts</summary>

https://github.com/vivid-planet/comet-starter/assets/42858722/89ea226f-e5e1-42d7-bc07-46526802b1cb

</details>

PR #275, opened 2024-06-25
Add `MediaBlock`

SebiVPS:
Add combined media block, and remove other media blocks.

The MediaBlock itself can be used in composite blocks, the aspect ratio should be handled.
The StandaloneMediaBlock is the standalone solution with aspect ratio selection.

- [ ] remove site implementation of YouTubeVideoBlock and DamVideoBlock and replace with comet blocks (https://github.com/vivid-planet/comet/pull/2223)


site/src/common/blocks/YouTubeVideoBlock.tsx
```diff
@@ -0,0 +1,86 @@
...
+};
+
+// TODO: use aspectRatio and previewImage from the block data when available
+interface YouTubeVideoBlockProps extends PropsWithData<YouTubeVideoBlockData> {
+    previewImage?: DamImageBlockData;
+    aspectRatio?: string;
+    sizes?: string;
+}
+
+export const YouTubeVideoBlock = withPreview(
```
nsams, line 26: i think we have this block in cms-site

nsams:
can you explain the sizes string?

site/src/common/blocks/DamImageBlock.tsx
```diff
@@ -1,33 +1,73 @@
...
+interface StaticLayout {
+    variant: "fixed" | "intrinsic";
+}
+
+type DamImageProps = Omit<NextImageProps, "src" | "width" | "height" | "layout" | "alt"> & {
+    aspectRatio?: string | "inherit";
+    layout?: DynamicLayout | StaticLayout;
+};
+
+const allowedAspectRatios = ["16x9", "4x3", "3x2", "3x1", "2x1", "1x1", "1x2", "1x3", "2x3", "3x4", "9x16", "21x9", "inherit"];
```
nsams, line 22: how is this list related to the one defined in comet.config?

SebiVPS:
> can you explain the sizes string?

@nsams here it is explained in detail: https://nextjs.org/docs/pages/api-reference/components/image#sizes
we use it to load the images in the expected resolution.

site/src/common/blocks/YouTubeVideoBlock.tsx
SebiVPS: we have one, but we do not want to use the required aspect ratio, this will be removed / made optional in the soon future

site/src/common/blocks/YouTubeVideoBlock.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+        if (autoplay !== undefined || (hasPreviewImage && !showPreviewImage))
+            searchParams.append("autoplay", Number(autoplay || (hasPreviewImage && !showPreviewImage)).toString());
+        if (autoplay) searchParams.append("mute", "1");
+
+        if (showControls !== undefined) searchParams.append("controls", Number(showControls).toString());
+
+        if (loop !== undefined) searchParams.append("loop", Number(loop).toString());
+        // the playlist parameter is needed so that the video loops. See https://developers.google.com/youtube/player_parameters#loop
+        if (loop && identifier) searchParams.append("playlist", identifier);
```
jamesricky, line 47: I'd still prefer ifs with {} but if you prefer without in this case, that's fine as well 😄

https://github.com/vivid-planet/comet-starter/pull/268#discussion_r1652688156

site/src/common/blocks/DamVideoBlock.tsx
```diff
@@ -1,18 +1,60 @@
...
+            <>
+                {hasPreviewImage && showPreviewImage && (
+                    <VideoPreviewImage
+                        onClick={() => setShowPreviewImage(false)}
+                        image={previewImage}
+                        aspectRatio={aspectRatio !== "auto" ? aspectRatio : undefined}
+                        sizes={sizes}
+                    />
+                )}
+                {(!showPreviewImage || !hasPreviewImage) && (
```
jamesricky, line 33: These conditions can be combined
https://github.com/vivid-planet/comet-starter/pull/268#discussion_r1652683960

```suggestion
                {hasPreviewImage && showPreviewImage ? (
                    <VideoPreviewImage
                        onClick={() => setShowPreviewImage(false)}
                        image={previewImage}
                        aspectRatio={aspectRatio !== "auto" ? aspectRatio : undefined}
                        sizes={sizes}
                    />
                ) : (
```

jamesricky (dismissed):

site/src/common/blocks/DamVideoBlock.tsx
SebiVPS: yep, thats true 👍

site/src/common/blocks/YouTubeVideoBlock.tsx
nsams: > this will be removed / made optional

you mean this copy in the starter will be removed?

> but we do not want to use the required aspect ratio

I don't understand, can you explain that in depth?

site/src/common/blocks/DamImageBlock.tsx
nsams: api does this based on aspect ratios defined in [comet-config.json](https://github.com/vivid-planet/comet-starter/blob/main/api/src/comet-config.json) - a config file shared across all microservices. Here you have a copy of those values that will get easily out of sync. I think we should reuse the values from comet-config.json.

site/src/common/blocks/DamImageBlock.tsx
```diff
@@ -1,33 +1,73 @@
...
+
+        aspectRatioNearestToAllowed = allowedAspectRatios[keyForNearestAspectRatio];
+    }
+    return aspectRatioNearestToAllowed;
+};
 
 export const DamImageBlock = withPreview(
-    ({ data: { block }, aspectRatio = "16x9", layout = "intrinsic", ...imageProps }: Props) => {
+    ({ data: { block }, aspectRatio = "16x9", layout, ...imageProps }: PropsWithData<DamImageBlockData> & DamImageProps) => {
+        const aspectRatioNearestToAllowed = getAspectRationNearestToAllowed(aspectRatio);
```
nsams, line 55: Is this really a good idea?
- Previous behavior: if you specify an unsupported aspectRatio the image won't load (correct me if I'm wrong), to fix this error either use an supported aspectRatio or extend the list of supported aspectRatios
- New behavior: if you specify an unsupported aspectRatio, it will be replaced by a hopefully close enough without giving any hints that this happend

site/src/common/blocks/DamImageBlock.tsx
SebiVPS: seems legit 👍 
is there a way to avoid the eslint-disable import/no-extraneous-dependencies ?
Screenshot 2024-06-26 at 08 39 16

site/src/common/blocks/YouTubeVideoBlock.tsx
SebiVPS: obi will explain it to you in your weekly 👍

site/src/common/blocks/DamImageBlock.tsx
```diff
@@ -42,3 +82,32 @@ export const DamImageBlock = withPreview(
     },
     { label: "Image" },
 );
+
+export const createSizes = (
```
nsams, line 88: what is the plan with this function?

admin/src/common/blocks/MediaBlock.tsx
```diff
@@ -0,0 +1,16 @@
+import { BlockCategory, BlockInterface, createOneOfBlock, YouTubeVideoBlock } from "@comet/blocks-admin";
+import { DamImageBlock, DamVideoBlock } from "@comet/cms-admin";
+import { FormattedMessage } from "react-intl";
+
+/*
+ * TODO: replace with new implementations with optional aspectRatio and video control options and preview images when available
```
nsams, line 6: are the plans somewhere documented, or can you explain them here?

site/src/common/blocks/DamImageBlock.tsx
```diff
@@ -1,33 +1,75 @@
...
+    sizes: string;
+}
+
+interface StaticLayout {
+    variant: "fixed" | "intrinsic";
+}
+
+type DamImageProps = Omit<NextImageProps, "src" | "width" | "height" | "layout" | "alt"> & {
+    aspectRatio?: string | "inherit";
+    layout?: DynamicLayout | StaticLayout;
```
nsams, line 21: why move that into layout?

site/src/common/blocks/DamImageBlock.tsx
SebiVPS: One could indeed question whether valid aspect ratios should be allowed at all. how did this actually come about?
@johnnyomair @mariokemetinger should we support this at all? How can it happen that an unsupported aspect ratio is used?

site/src/common/blocks/DamImageBlock.tsx
SebiVPS: readability & usability 👍

nsams: I'm asking because imho we should stay as close as possible to next Image.

Note: that [changed](https://nextjs.org/docs/pages/api-reference/components/image-legacy#comparison) quite a bit in the new nextjs version (which we are not yet using in main)

SebiVPS: Ok, i tried to remove `layout` and `sizes` prop and use them from next/image but the PixelImageBlock does use its own layout prop `Types of property layout are incompatible.`. https://github.com/vivid-planet/comet/blob/main/packages/site/cms-site/src/blocks/PixelImageBlock.tsx

When we want to stay as close as possible to nextjs image we should change that too.

SebiVPS: Maybe we can improve usage of the DamImageBlock in another way?
When you do not set `layout={"responsive"} sizes={sizes}` a full resolution dam image (~4000px) is loaded which is not really performant in most use cases.

nsams: > When we want to stay as close as possible to nextjs image we should change that too.

ok, I wasn't aware of that. That has room for improvements, however that should be done in next branch, for the new nextjs image component.

> a full resolution dam image (~4000px) is loaded which is not really performant in most use cases.

not good. But isn't that a matter of setting sizes correctly?

SebiVPS: > a full resolution dam image (~4000px) is loaded which is not really performant in most use cases.
> 
> not good. But isn't that a matter of setting sizes correctly?

Yes it is.

site/src/common/blocks/DamImageBlock.tsx
```diff
@@ -1,33 +1,64 @@
...
-        aspectRatio?: string;
-    } & (
-        | { layout?: "fixed" | "intrinsic" }
-        // The sizes prop must be specified for images with layout "fill" or "responsive", as recommended in the next/image documentation
-        // https://nextjs.org/docs/api-reference/next/image#sizes
-        | {
-              layout?: "fill" | "responsive";
-              sizes: string;
-          }
-    );
```
SebiVPS, line 19: @nsams just found the root of this code: https://github.com/vivid-planet/comet/blob/d77237de93d823243a78bcc43c49870bdb0ad89d/packages/site/cms-site/src/image/Image.tsx#L64 

the wrapping in DynamicLayout / StaticLayout was a try of someone to make this more readable i guess.
but as you mention, we should stay close to nextjs image definition, it is redundant at the end.

site/src/common/blocks/DamImageBlock.tsx
SebiVPS: i found this function in all newer projects and i think it will be needed. The best use case i can imagine is an image in the columns block: When you have 2 columns layout you want to load an image in large layout with eg. 50vw max width, but on the small phone layout the 2 column layout transforms into 1 column so you have a full width image with 100vw. To fulfill this behaviour you need the sizes, with this helper function this sizes string is created for you.
`sizes={createSizes({ sx: "100vw", sm: "100vw", md: "50vw", lg: "50vw" })}`

site/src/common/blocks/DamImageBlock.tsx
SebiVPS: i removed this implementation for now, we can discuss if we really need this later on

site/src/common/blocks/DamImageBlock.tsx
nsams: I understand. Have you considered adding support for breakpoints in the sizes prop of the block component? That would even allow getting the theme from context (instead of importing the theme like you do now)

SebiVPS: Sorry, i do not understand what you mean. Can you please give an example what you suggest?

SebiVPS:
closed due to next migration #296

johnnyomair:
For the next time: Rebasing onto next would also have been an option. Then the previous conversations wouldn't be lost.

SebiVPS:
I thought merging is the lesser evil, but that did not work out either, as you see the changes here. Maybe the merge main make too much conflicts.

PR #280, opened 2024-06-26
Update/Add security headers

max-debug022:
This PR improves/updates the HTTP-Headers

---
**TLDR**
- Update helmet to 7.2.0 in API and Admin
- Coherent CORS-Policy (Cross-Origin-Ressource-Sharing) for the API -> to Admin & Site
- Coherent CORP/COEP (Cross-Origin-Ressource/Embedder-Policy) for API, Admin & Site
- Strict COOP (Cross-Origin-Opener-Policy) for API, Admin & Site
- Removal of deprecated, redundant, or prohibited headers
- CSP (Content Security Policy) that enforces the JEA (just enough access) principle 
<details>
This PR implements the "Just-Enough-Access" (JEA) security strategy, ensuring that only necessary privileges are granted while restricting all others. The PR includes a coherent CORS policy so that resources can only be shared within the microservices. Additionally, CORP/COEP have been enforced to control resource access and embedding to authorized origins only. Several Content Security Policy (CSP) errors have been resolved & unnecessary privileges have been removed - Admin has slightly more flexibility than the Site. Reasons for several changes:

- Update helmet to v7.1.0
- CORS uses now a maxAge of 600s as Chromium (prior to v76) caps at 10mins 
- CORP only for Same-Site
- Set COOP for same-origin
- Remove the empty exposedHeaders field as it is the default
- Disable unused x-powered-by header
- Disable [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection) X-XSS-Protection Header
- Disable [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) X-Frame-Options Header
- Enable HSTS with recommended values 
- Disable Referrer Policy 
- ...

</details>


thomasdax98:
Did you test this in one of our projects in a deployed setting? I fear that there might be some unwanted side-effects relating to different domains in a deployed setting

max-debug022:
> Did you test this in one of our projects in a deployed setting? I fear that there might be some unwanted side-effects relating to different domains in a deployed setting

Yes, I tested it on deployed settings.

dkarnutsch:
> Did you test this in one of our projects in a deployed setting? I fear that there might be some unwanted side-effects relating to different domains in a deployed setting

I would not add this to existing projects without extensive testing. However, the starter can define best practices and hopefully all bugs are discovered during development. So I think it is relatively safe to merge this.

dkarnutsch:
Some header could possible have a comment, when to extend/change them.

max-debug022:
> Some header could possible have a comment, when to extend/change them.
> 
> You could also add links to best practices (MDN?), so we can track changes.

Luckily I originally added comments to each Header but removed them since I wasn't sure if comments should be in the code. I could add them, but I think it would be better to add detailed docs on how to modify each header and what effect each header has. What would you prefer?

dkarnutsch:
> > Some header could possible have a comment, when to extend/change them.
> > You could also add links to best practices (MDN?), so we can track changes.
> 
> Luckily I originally added comments to each Header but removed them since I wasn't sure if comments should be in the code. I could add them again, but I think it would be better to add detailed docs on how to modify each header and what effect each header has. What would you prefer?

In code, as documentation can drift away. @johnnyomair?

johnnyomair:
> > > Some header could possible have a comment, when to extend/change them.
> > > You could also add links to best practices (MDN?), so we can track changes.
> > 
> > 
> > Luckily I originally added comments to each Header but removed them since I wasn't sure if comments should be in the code. I could add them again, but I think it would be better to add detailed docs on how to modify each header and what effect each header has. What would you prefer?
> 
> In code, as documentation can drift away. @johnnyomair?

Agreed. We could add line comments for each header/item.

site/next.config.js
```diff
@@ -86,30 +109,11 @@ const nextConfig = {
...
-                            `
-                        .replace(/\s{2,}/g, " ")
-                        .trim(),
+                    value: "same-origin",
                 },
                 {
                     key: "Access-Control-Allow-Origin",
-                    value: process.env.ADMIN_URL,
+                    value: "none",
                 },
```
johnnyomair, line 117: Why is it okay to remove this?

site/next.config.js
```diff
@@ -76,6 +89,16 @@ const nextConfig = {
                     key: "Cross-Origin-Opener-Policy",
                     value: "same-origin",
                 },
+                {
+                    key: "Cross-Origin-Embedder-Policy",
+                    // This value should be set to "require-corp" as soon as iframe credentialless it is supported by all browsers
+                    // https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless
```
thomasdax98, line 95:
```suggestion
                    // This value should be set to 'require-corp' as soon as iframe credentialless is supported by all browsers
                    // https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless
                    // https://caniuse.com/mdn-html_elements_iframe_credentialless
```

site/next.config.js
thomasdax98: Is "none" even a valid value? It's not mentioned on MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

site/next.config.js
```diff
@@ -76,6 +89,16 @@ const nextConfig = {
...
+                    value: "unsafe-none",
+                },
+                {
+                    key: "Cross-Origin-Resource-Policy",
+                    value: "same-site",
```
thomasdax98, line 100: Why `same-site` instead of `same-origin`?

api/src/main.ts
```diff
@@ -46,9 +47,34 @@ async function bootstrap(): Promise<void> {
...
+                policy: "no-referrer",
+            },
+            xContentTypeOptions: true,
+            xDnsPrefetchControl: false,
+            xDownloadOptions: true,
+            xPermittedCrossDomainPolicies: true,
+            originAgentCluster: true,
+            crossOriginResourcePolicy: {
+                policy: "same-site",
+            },
```
thomasdax98, line 77: Doesn't this limit the access to the API to our API domain? 

I believe we have some client projects where at least API and site use different domains

site/next.config.js
thomasdax98: Ah I guess it's because same-site allows different sub domains

site/next.config.js
max-debug022: @johnnyomair  Since this site should only be embedded as an iframe, focusing on CORP/COEP makes sense for security, and it doesn’t need to handle cross-origin requests. Therefore, removing the admin url in Access-Control-Allow-Origin should be ok because CORS mainly hanldes cross-origin data access, typically for APIs.

@thomasdax98 Given that ACAO is unnecessary, we have three options: omit it, set it to an invalid value, or use "none". I prefer setting it to "none" or omitting it entirely. This is because the default behavior for ACAO in Next.js isn’t explicitly documented, and relying on defaults that could change over time isn’t ideal imho.

admin/server/server.js
```diff
@@ -7,24 +7,43 @@ const app = express();
...
+                "font-src": ["'self'", "data:"],
+                "connect-src": ["https:"],
+                "img-src": ["'self'", "data:", "https:"],
+                "frame-src": ["https:"],
+                upgradeInsecureRequests: [],
             },
+            upgradeInsecureRequests: true,
+            useDefaults: false,
         },
         xXssProtection: false,
```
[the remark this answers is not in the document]
max-debug022: It’s just a personal preference to explicitly set values even if they are implicitly set by default. This way, anyone looking at the code doesn't need to refer to the documentation to understand the defaults. Plus, if the default behavior ever changes, our explicitly set values ensure our application’s behavior remains the same.

admin/server/server.js
```diff
@@ -7,24 +7,43 @@ const app = express();
...
+        xFrameOptions: false,
```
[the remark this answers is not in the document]
max-debug022: Because it is obsoleted by frame-ancestors: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

But since frame-ancestors has no fallback, I think we should specify this directive and set it to none or self.

admin/server/server.js
```diff
@@ -7,24 +7,43 @@ const app = express();
...
             maxAge: 63072000,
             includeSubDomains: true,
             preload: true,
         },
+        referrerPolicy: {
+            policy: "no-referrer",
+        },
+        xContentTypeOptions: true,
+        xDnsPrefetchControl: false,
+        xDownloadOptions: true,
```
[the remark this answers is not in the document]
max-debug022: Changed in: https://github.com/vivid-planet/comet-starter/pull/280#discussion_r1668052908

admin/server/server.js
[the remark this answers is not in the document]
max-debug022: https://github.com/vivid-planet/comet-starter/pull/280/commits/3221d1ac1eadf4a7e48a8fc85565e258b946234f

api/src/app.module.ts
```diff
@@ -67,7 +67,7 @@ export class AppModule {
                             origin: config.corsAllowedOrigin,
                             methods: ["GET", "POST"],
                             credentials: false,
-                            exposedHeaders: [],
+                            maxAge: 600,
```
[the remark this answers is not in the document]
max-debug022: 1) Reduced Preflight Requests
2) Improved performance by caching preflight requests

site/next.config.js
```diff
@@ -68,6 +68,19 @@ const nextConfig = {
...
             headers: [
+                {
+                    key: "Content-Security-Policy",
+                    value: [
+                        "default-src 'none'",
+                        "script-src-elem 'self'",
+                        "style-src-elem 'unsafe-inline'",
+                        "style-src-attr 'unsafe-inline'",
+                        `img-src 'self' data: ${process.env.API_URL}/`,
+                        "script-src 'unsafe-eval'",
```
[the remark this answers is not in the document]
max-debug022: Unfortunately, it seems like it is required by one of our dependencies.

site/next.config.js
max-debug022: Subdomains but most importantly ports.

api/src/main.ts
max-debug022: > Doesn't this limit the access to the API to our API domain?
Exactly

> I believe we have some client projects where at least API and site use different domains
We would need to change it in the client project.

max-debug022:
> Agreed. We could add line comments for each header/item.

@johnnyomair what headers should be documented and how? I don't think an explanation is useful since anyone can look headers up and they are fairly well explained. And if you understand the header you probably understand the logic behind their value. 

For example. x-xss-protection - the first thing in the MDN docs is that it should not be used, so its value is false.

admin/server/server.js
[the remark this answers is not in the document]
thomasdax98: > This way, anyone looking at the code doesn't need to refer to the documentation to understand the defaults.

Especially in this case, I'd say it's confusing for someone who isn't very familiar with the headers. XSS protection sounds like a good thing, so people might wonder why we are disabling it. It's not implicit that this header is actually bad. So this will prompt people to consult the docs even more. 

> Plus, if the default behavior ever changes, our explicitly set values ensure our application’s behavior remains the same.

Since this is an outdated, non-standard, harmful header, it's not a realistic scenario that the default ever changes to true. If anything, the header will be removed. While I'm normally also a fan of explicit configuration, I wouldn't say this really applies here.

But let's see what the others think about that

thomasdax98:
> Edit: My initial documentation was somewhat like this:
> 
> ```
> app.use(
>   helmet({
>     contentSecurityPolicy: {
>       directives: {
>         "default-src": ["'none'"], // No external resources are needed
>       },
>       useDefaults: false, // Disable all other default CSP rules
>     },
>     xXssProtection: false, // Disable deprecated header
>     xFrameOptions: false, // Disable deprecated header
>     crossOriginResourcePolicy: "same-origin", // Do not allow cross-origin requests to access the response
>     crossOriginEmbedderPolicy: true, // =no-corp Enable COEP
>     crossOriginOpenerPolicy: true, // =same-origin Enable COOP
>     strictTransportSecurity: {
>       // Enable HSTS
>       maxAge: 63072000, // 2 years (recommended if including subdomains)
>       includeSubDomains: true, // Include all subdomains
>       preload: true, // Enable HSTS preload
>     },
>     referrerPolicy: {
>       policy: "no-referrer", // No referrer information needs to be sent
>     },
>     xContentTypeOptions: true, // = nosniff
>     xDnsPrefetchControl: false, // Disable non-standard header as recommended by MDN
>     xDownloadOptions: true, // = noopen (forces IE8+ not to open files in the context of the page but instead download them)
>     xPermittedCrossDomainPolicies: true, // = none (prevent the browser from MIME-sniffing)
>     originAgentCluster: true, // = false (disables the agent cluster)
>   }),
> );
> ```

@max-debug022 I like this. During the review I had a hard time wrapping my head around "What does `true` (or `false`) mean in this case?"

I think you should add these comments

max-debug022:
I added the comments: https://github.com/vivid-planet/comet-starter/pull/280/commits/cbc047256c8285d648e77fa165b158a0abd119cb

admin/server/server.js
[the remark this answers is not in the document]
max-debug022: I noticed that false (= not send the header) is not the default value - "0" is. So removed the header since the default value of 0 is actually better than not setting it at all

thomasdax98 (approved):

site/next.config.js
johnnyomair: @max-debug022 but is "none" a valid value?

site/next.config.js
johnnyomair, line 1: As discussed, please also disable x-powered-by here

johnnyomair (approved):

max-debug022:
@thomasdax98 @fraxachun @thomasdax98 @johnnyomair

admin/server/server.js
```diff
@@ -7,24 +7,42 @@ const app = express();
...
+                "default-src": ["'none'"],
+                "script-src-elem": ["'unsafe-inline'", "'self'"],
+                "style-src-elem": ["'unsafe-inline'", "'self'"],
+                "style-src-attr": ["'unsafe-inline'"],
+                "font-src": ["'self'", "data:"],
+                "connect-src": ["https:"],
+                "img-src": ["'self'", "data:", "https:"],
+                "frame-src": ["https:"],
+                "frame-ancestors": ["'self'"],
+                upgradeInsecureRequests: [],
```
dkarnutsch, line 26: this is not consistent with the site (where this prop is missing),
also: this prop is not working with our domain setup (dani.dev.vivid-planet.cloud)

max-debug022: Added with: https://github.com/vivid-planet/comet-starter/pull/280/commits/31542001b14f22273959b9c30dc9db30fe87c471

site/next.config.js
```diff
@@ -24,11 +24,26 @@ const nextConfig = {
...
+                        "connect-src 'self'",
+                        "frame-src https://www.youtube-nocookie.com",
+                        "upgrade-insecure-requests", // Remove this option when working with the Domain-Setup
```
johnnyomair, line 44: This should be done automatically based on the presence of the environment variable.

johnnyomair: I guess you'll need the same fix in the Admin and API, right?

max-debug022: Yes: https://github.com/vivid-planet/comet-starter/pull/280/commits/7afbbf331528aea4ef10069e4625d7ce4f180fb8

site/next.config.js
```diff
@@ -24,11 +24,25 @@ const nextConfig = {
...
+                    key: "Content-Security-Policy",
+                    value: [
+                        "default-src 'none'",
+                        "script-src-elem 'self' 'unsafe-inline'",
+                        "style-src-elem 'unsafe-inline'",
+                        "style-src-attr 'unsafe-inline'",
+                        `img-src 'self' data: ${process.env.API_URL}/`,
+                        "script-src 'unsafe-eval'",
+                        "connect-src 'self'",
+                        `frame-src https://www.youtube-nocookie.com; ${!process.env.DEV_DOMAIN ? "upgrade-insecure-requests" : ""}`, // Don't use upgrade-insecure-requests with the Domain-Setup
```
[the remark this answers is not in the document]
max-debug022: https://github.com/vivid-planet/comet-starter/pull/280/commits/7afbbf331528aea4ef10069e4625d7ce4f180fb8

johnnyomair:
@max-debug022 please fix conflicts.

site/next.config.js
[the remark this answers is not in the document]
max-debug022: Well, it's a lot; and I do not know an efficient way to evaluate it (even with tools like bundle-analyzer). Just because a package contains eval or any other "unsafe-eval" code doesn't necessarily mean the CSP needs to permit it. We have hundreds of files (including package dependencies oc) that include some form of eval.

A few example-deps are:
- exceljs
- protobufjs
- next.js
- terser-webpack-plugin
- uglify-js

There are many others, particularly code minifiers, TypeScript dependencies, and linting libraries. Has anyone suggestions on a more effective way to evaluate this?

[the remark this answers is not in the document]
johnnyomair: > Has anyone suggestions on a more effective way to evaluate this?

I'd probably set it to 'none', start the site and see where an error occurs.

> exceljs

I doubt that we have this dependency in the site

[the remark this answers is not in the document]
max-debug022: > I'd probably set it to 'none', start the site and see where an error occurs.
That's what I did in the first place, but the error messages are generic and rarely inform you about the specific package.
> I doubt that we have this dependency in the site
Yes, sorry - I forgot to mention that I included all microservices since we not only use the CSP in the site, (although API wouldn't be necessary I guess).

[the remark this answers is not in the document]
johnnyomair: We can have a look at it together the next time we're both in the office.

site/next.config.js
```diff
@@ -24,11 +24,25 @@ const nextConfig = {
...
             source: "/:path*",
             headers: [
+                {
+                    key: "Content-Security-Policy",
+                    value: [
+                        "default-src 'none'",
+                        "script-src-elem 'self' 'unsafe-inline'",
+                        "style-src-elem 'unsafe-inline'",
+                        "style-src-attr 'unsafe-inline'",
+                        `img-src 'self' data: ${process.env.API_URL}/`,
```
johnnyomair, line 40: I believe we don't need the API_URL here anymore since we're using a rewrite to proxy the images through the site: #298

johnnyomair (changes requested):
Please fix the lint workflow

johnnyomair (dismissed):

johnnyomair:
@dkarnutsch @fraxachun @thomasdax98 please review, this one's been open for way too long 😁

site/next.config.js
```diff
@@ -24,50 +24,59 @@ const nextConfig = {
...
-                                default-src 'self';
-                                form-action 'self'; 
-                                object-src 'none';
-                                img-src 'self' https: data:${process.env.NODE_ENV === "development" ? " http:" : ""};
-                                media-src 'self' https: data:${process.env.NODE_ENV === "development" ? " http:" : ""};
-                                style-src 'self' 'unsafe-inline'; 
-                                font-src 'self' https: data:;
-                                script-src 'self' 'unsafe-inline' https:${process.env.NODE_ENV === "development" ? " 'unsafe-eval'" : ""};
-                                connect-src 'self' https:${process.env.NODE_ENV === "development" ? " http:" : ""};
-                                frame-ancestors ${process.env.ADMIN_URL};
```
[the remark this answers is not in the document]
max-debug022: Yes thats true: https://github.com/vivid-planet/comet-starter/pull/280/commits/e7050c55f36d686e059540befcf4e662c15dab71

site/next.config.js
```diff
@@ -24,50 +24,59 @@ const nextConfig = {
...
+                        "default-src 'self'", // Needed for svgs to work in Firefox (other browsers load svgs with img-src)
+                        "style-src-elem 'self' 'unsafe-inline'",
+                        "style-src-attr 'unsafe-inline'",
+                        "script-src-elem 'self' 'unsafe-inline'",
+                        `${process.env.NODE_ENV === "development" ? "script-src 'unsafe-eval'" : ""}`, // Needed in local development
+                        `${process.env.NODE_ENV === "development" ? "connect-src ws:" : ""}`, // Used for hot reloading in local development
+                        "font-src data:",
+                        "frame-src https://www.youtube-nocookie.com/",
+                        `img-src data: 'self' ${process.env.NODE_ENV === "development" ? process.env.API_URL + "/" : ""}`,
+                        `${process.env.NODE_ENV === "development" ? "" : "upgrade-insecure-requests"}`, // Don't use upgrade-insecure-requests with the Domain-Setup
```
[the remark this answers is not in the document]
max-debug022: What do you think about this: https://github.com/vivid-planet/comet-starter/pull/280/commits/42a5ebfdd8c5ea5aff990f0fb72124c5336a6151

site/next.config.js
```diff
@@ -24,50 +24,59 @@ const nextConfig = {
...
+                    value: [
+                        "default-src 'self'", // Needed for svgs to work in Firefox (other browsers load svgs with img-src)
+                        "style-src-elem 'self' 'unsafe-inline'",
+                        "style-src-attr 'unsafe-inline'",
+                        "script-src-elem 'self' 'unsafe-inline'",
+                        `${process.env.NODE_ENV === "development" ? "script-src 'unsafe-eval'" : ""}`, // Needed in local development
+                        `${process.env.NODE_ENV === "development" ? "connect-src ws:" : ""}`, // Used for hot reloading in local development
+                        "font-src data:",
+                        "frame-src https://www.youtube-nocookie.com/",
+                        `img-src data: 'self' ${process.env.NODE_ENV === "development" ? process.env.API_URL + "/" : ""}`,
```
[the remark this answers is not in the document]
max-debug022: Fixed in: https://github.com/vivid-planet/comet-starter/pull/280/commits/42a5ebfdd8c5ea5aff990f0fb72124c5336a6151

site/next.config.js
```diff
@@ -5,6 +5,31 @@ const withBundleAnalyzer = require("@next/bundle-analyzer")({
     enabled: process.env.ANALYZE === "true",
 });
 
+function generateCSP() {
```
johnnyomair, line 8: headers already is a function, so why not move the code there before returning the array?

site/next.config.js
```diff
@@ -5,6 +5,31 @@ const withBundleAnalyzer = require("@next/bundle-analyzer")({
...
+        "img-src": `data: 'self' ${process.env.API_URL}`,
+        "frame-ancestors": process.env.ADMIN_URL,
+    };
+
+    // Conditionally add environment-specific rules
+    if (process.env.NODE_ENV === "development") {
+        cspRules["script-src"] = "'unsafe-eval'"; // Needed in local development
+        cspRules["connect-src"] = "ws:"; // Used for hot reloading in local development
+    } else {
+        cspRules["upgrade-insecure-requests"] = ""; // Don't use upgrade-insecure-requests with Domain-Setup
```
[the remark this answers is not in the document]
max-debug022: This instructs user agents to treat a site's insecure URLs (HTTP) as though they have been replaced with secure URLs.

max-debug022:
I will wait for this PR to be merged: https://github.com/vivid-planet/comet-starter/pull/415

site/next.config.js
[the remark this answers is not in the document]
max-debug022: Including upgrade-insecure-requests is optional but helps avoid potential misconfigurations and adds an extra layer of security, so I’d recommend keeping it.

thomasdax98 (dismissed):

api/package.json
```diff
@@ -52,8 +52,7 @@
         "compression": "^1.7.5",
         "cookie-parser": "^1.4.7",
         "express": "^4.21.2",
-        "graphql": "^15.10.1",
```
dkarnutsch, line 55: why? this seems unrelated to this PR

johnnyomair: Also, this shouldn't be removed.

api/project-words.txt
johnnyomair, line 1: Please remove this file

site/next.config.mjs
```diff
@@ -25,30 +25,44 @@ const nextConfig = {
...
+                    // https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless
+                    // https://caniuse.com/mdn-html_elements_iframe_credentialless
+                    value: "unsafe-none",
+                },
+                {
+                    key: "Cross-Origin-Resource-Policy", // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Resource-Policy
+                    value: "same-site", // Do not allow cross-origin requests to access the response
+                },
+                {
+                    //
```
johnnyomair, line 54:
```suggestion
```

johnnyomair (changes requested):

johnnyomair (approved):

site/src/middleware/contentSecurityPolicyHeaders.ts
```diff
@@ -2,29 +2,36 @@ import { NextRequest } from "next/server";
...
+type ContentSecurityPolicyDirective = {
+    directive: string;
+    values: string[];
+};
+
+const contentSecurityPolicyDirectives: ContentSecurityPolicyDirective[] = [
+    { directive: "default-src", values: ["'none'"] }, // Restrict any content not explicitly allowed
+    { directive: "style-src-elem", values: ["'self'", "'unsafe-inline'"] },
+    { directive: "script-src-elem", values: ["'self'", "'unsafe-inline'"] },
+    { directive: "img-src", values: ["'self'", "data:", process.env.API_URL?.replace("/api", "") ?? ""] },
```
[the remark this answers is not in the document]
johnnyomair: @max-debug022 can we remove the API_URL? The string replacement of /api is odd IMO.

johnnyomair:
> Shouldn't it be comet-demo first?

This PR has been started long before we changed to demo first, so I'd like to finish it here and then copy it to Demo

dkarnutsch (approved):

site/src/middleware/contentSecurityPolicyHeaders.ts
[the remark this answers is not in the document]
max-debug022: We need this for images in the block-preview iframe:
image
image

thomasdax98 (approved):

PR #283, opened 2024-07-04
Use site configs in site

magdalenaxm:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-988


site/src/middleware.ts
```diff
@@ -21,7 +41,11 @@ export async function middleware(request: NextRequest) {
         return NextResponse.rewrite(new URL(rewrite.destination, request.url));
     }
 
-    return NextResponse.next();
+    const response = NextResponse.next({ request: { headers } });
+
+    // TODO are there any other headers that need to be set?
```
magdalenaxm, line 46: I'm not quite sure how TODO's are handled in this project, but in the reference project there is another set of headers. But when adding them here, the block preview no longer works.

johnnyomair: Could you give an example which headers are set?

johnnyomair (commented):
As discussed, please change the site configs so that we have one config for each domain with multiple languages.

admin/src/documents/pages/EditPage.tsx
```diff
@@ -129,7 +129,7 @@ export const EditPage: React.FC<Props> = ({ id, category }) => {
                             startIcon={<Preview />}
                             disabled={!pageState}
                             onClick={() => {
-                                openSitePreviewWindow(pageState.path, contentScopeMatch.url);
+                                openSitePreviewWindow(`/${scope.language}${pageState.path}`, contentScopeMatch.url);
```
thomasdax98, line 132: I decided to add the language to the path in admin for simplicity. This could alternatively be done in the redirect in `site/api/site-preview/route.ts` or directly in site `page.tsx`.

thomasdax98: Update: This doesn't work because it results in a wrong link in the site preview. To resolve this, a workaround in the lib would be needed which isn't a nice solution

Bildschirmfoto 2024-07-12 um 11 40 12

site/src/temp/site-preview-route.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    const data: SitePreviewParams = { scope, previewData };
+    const token = await new SignJWT(data).setProtectedHeader({ alg: "HS256" }).sign(new TextEncoder().encode(previewScopeSigningKey));
+    cookies().set("__comet_preview", token);
+
+    draftMode().enable();
+
+    return redirect(`/${path.join(options?.basePath?.(request) ?? "", params.get("path") ?? "")}` || "/");
+}
```
thomasdax98, line 68: Note: This is a copy of https://github.com/vivid-planet/comet/blob/next/packages/site/cms-site/src/sitePreview/SitePreviewUtils.ts#L28

I added the basePath option to enable putting parts of the scope in the URL. I wanted to try if the approach works before implementing it in the library

site/src/middleware.ts
```diff
@@ -2,10 +2,38 @@ import { Rewrite } from "next/dist/lib/load-custom-routes";
...
 export async function middleware(request: NextRequest) {
+    const headers = request.headers;
+    const host = getHost(headers);
     const { pathname } = new URL(request.url);
+    const language = pathname.split("/")?.[1] ?? "";
+
+    if (!language) {
+        return NextResponse.redirect(new URL("/en", request.url));
+    }
+    headers.set("x-current-language", language);
```
nsams, line 18: Why not access the lang param, you have that also in all sub pages. (not in siteConfig() tough, you would have to pass lang into it i guess)

site-configs.ts
```diff
@@ -3,31 +3,50 @@ import { SiteConfig } from "./site-configs.d";
 
 // Types for files in site-configs/
 type Environment = "local" | "dev" | "test" | "staging" | "prod";
-export type Config = Omit<SiteConfig, "domains"> & {
+export type Config = Omit<SiteConfig, "domains" | "contentScope"> & {
+    contentScope: {
+        domain: string;
+        languages: string[];
```
nsams, line 9: hm, languages is never part of the contentScope, only language is. Maybe this should be moved one level up in the config

site-configs.ts
```diff
@@ -3,31 +3,50 @@ import { SiteConfig } from "./site-configs.d";
...
+        const ret = languages.map((language):SiteConfig => {
+            return {
+                ...site,
+                name: `${site.name} ${language.toUpperCase()}`,
+                contentScope: {
+                    domain: site.contentScope.domain,
+                    language
+                },
+                domains: {
+                    main: `${domains[env]}/${language}` ?? "",
```
nsams, line 41: this is now a domain including a path.... I thought more about one siteConfig for two languages. (But I haven't though about all details)

site/src/middleware.ts
thomasdax98: That works for `site/src/app/[lang]/[[...path]]/layout.tsx` and `site/src/app/[lang]/[[...path]]/page.tsx`. 

But it doesn't work for the `RootLayout` in `site/src/app/layout.tsx` and `sitemap` in `site/src/app/[lang]/sitemap.ts`. 

Theoretically, I could make `RootLayout` a client component and access the params via `useRouter` (which isn't a good idea). And this approach doesn't work for `sitemap` since it's not a React component.

site-configs.ts
thomasdax98: Yes, we could do that. I wasn't sure about the best position

site-configs.ts
thomasdax98: So you mean leaving the domain as is and keeping the language in an extra field. Then extracting the language from the URL and comparing it separately from the domain?

site/src/middleware.ts
nsams: ### RootLayout

you should be able to move everything langauge dependent into `src/app/[lang]` (or `src/app/[lang]/[[...path]]`). See demo for IntlProvider.

### sitemap

I think there should be a `www.example.com/sitemap.xml` (not a `www.example.com/en/sitemap.xml` and `www.example.com/de/sitemap.xml`) containing urls from all langauges. (see demo)

[Splitting up sitemap](https://developers.google.com/search/docs/crawling-indexing/sitemaps/large-sitemaps) into multiple urls (per domain) is usually done for performance resons only (if it gets too large).

site-configs.ts
nsams: mainly I'm concered about calling it "domain" but beeing domain+path.

If you use a siteConfig per domain (not domain+language) you don't have this anymore. I would really try that.

johnnyomair:

site/src/middleware.ts
thomasdax98: Done in https://github.com/vivid-planet/comet-starter/pull/283/commits/814f5826fa6f6c181ce9ea4054c171f97c6ee6c8

site-configs.ts
thomasdax98: Done in https://github.com/vivid-planet/comet-starter/pull/283/commits/f5ffb6254405ad9e328e5193d897d7ae8b6b09fd

site-configs.ts
thomasdax98: I implemented another version where the site config is shared for languages (so one site config per domain): https://github.com/vivid-planet/comet-starter/pull/283/commits/a7ec7377ba5278a05e1ee7f45f4ee5926e95bf6b

What do you think? 

The typing is still somewhat hacky. If we want to go with this approach, I think we should adjust the types in the library

admin/src/App.tsx
```diff
@@ -57,13 +57,16 @@ export function App() {
...
+
                                 if (!siteConfig) throw new Error(`siteConfig not found for domain ${scope.domain}`);
                                 return {
-                                    url: siteConfig.url,
+                                    url: `${siteConfig.url}/${scope.language}`,
                                     preloginEnabled: siteConfig.preloginEnabled || false,
-                                    blockPreviewBaseUrl: `${config.previewUrl}/block-preview`,
-                                    sitePreviewApiUrl: `${config.previewUrl}/api/site-preview`,
+                                    blockPreviewBaseUrl: `${siteConfig.public.previewUrl}/block-preview`,
+                                    sitePreviewApiUrl: `${siteConfig.public.previewUrl}/${scope.language}/api/site-preview`,
```
nsams, line 69: I don't think there is a reason why you should put the api route under [lang], just move it one level up

site/src/config.ts
```diff
@@ -1,3 +1,33 @@
...
+            siteConfig = getSiteConfigFromScope(preview.scope as ContentScope);
+        }
+    }
+
+    if (!siteConfig) notFound();
+    return siteConfig;
+}
+
+export function getHost(headers?: Headers) {
+    if (typeof window !== "undefined" && window.location?.host) return window.location.host;
```
nsams, line 27: what is the use case for this function to support in server side and client side?

admin/src/App.tsx
thomasdax98: This was suggested by @fraxachun so we can automatically determine the `basePath` in `sitePreviewRoute` (see https://github.com/vivid-planet/comet-starter/pull/283/commits/21e9ad6e4a7f97000b3305b7fec54af8c7af639b#diff-5f1f3b4b538a3239e83006262d8380217ad7e1214d4a6dc485ea52417993ffcfR68)

The alternative would be a `basePath` function to extract the `basePath` from the query params (as done in https://github.com/vivid-planet/comet-starter/pull/283/commits/93b35a450f0151efb540eeb409807692fab31104#diff-eeaf5aa604eda74648fc8e3d10fec1fa195b906ef4fb9ffc947becbd6d5fa037R9-R18)

What do you prefer?

site/src/temp/site-preview-route.ts
```diff
@@ -0,0 +1,71 @@
...
+    const token = await new SignJWT(data).setProtectedHeader({ alg: "HS256" }).sign(new TextEncoder().encode(previewScopeSigningKey));
+    cookies().set("__comet_preview", token);
+
+    draftMode().enable();
+
+    const apiRoutePostfix = options?.apiRoutePostfix ?? "/api/site-preview";
+    const basePath = request.nextUrl.pathname.split(apiRoutePostfix)[0];
+
+    return redirect(path.join(basePath, params.get("path") ?? "") || "/");
+}
```
thomasdax98, line 71: This change is made in the core library in https://github.com/vivid-planet/comet/pull/2303

admin/src/App.tsx
nsams: https://github.com/vivid-planet/comet/pull/2303#issuecomment-2238326538

thomasdax98: I just realised that you already solved this problem in Demo using `SitePreview#resolvePath` (https://github.com/vivid-planet/comet/blob/979eb54f71aa37ed1467ce5d2f666d0b4edc14be/demo/admin/src/App.tsx#L124-L126). 

I further realised, that the `SitePreviewProvider` was still missing here in the Starter, so I added it (https://github.com/vivid-planet/comet-starter/pull/283/commits/669cd1233e04a9ab554c55c33300361c5ab3b310). 

My solution doesn't work together with `SitePreviewProvider` because it has a mechanism built in to send the site path back to the admin (https://github.com/vivid-planet/comet/blob/52d32febef89b266810de783b78f4bc8782728df/packages/site/cms-site/src/sitePreview/SitePreviewProvider.tsx#L15-L19). This means my rewrite in the API route is in turn overwritten resulting in multiple language prefixes in the path -> it doesn't work.

So I had two options:

1. Either do it like it was done in the Demo
2. Change `SitePreviewProvider` and `sitePreviewRoute` in the library to use my solution

I opted for option 1 and now we have the same solution as in the Demo: a64c844fbd4c00a6c8330a3054febfd8468ad2fa

This requires no change in the library and we can close https://github.com/vivid-planet/comet/pull/2303

site/src/middleware.ts
```diff
@@ -2,10 +2,34 @@ import { Rewrite } from "next/dist/lib/load-custom-routes";
...
+import { getSiteConfigs } from "./siteConfigs";
 
 export async function middleware(request: NextRequest) {
+    const headers = request.headers;
+    const host = getHost(headers);
     const { pathname } = new URL(request.url);
+    const language = pathname.split("/")?.[1] ?? "";
+
+    if (!language) {
+        return NextResponse.redirect(new URL("/en", request.url));
```
nsams, line 16: I prefer the implementation from demo: https://github.com/vivid-planet/comet/blob/next/demo/site/src/app/page.tsx

- is at the imho logical location
- doesn't need to parse the url for language
- doesn't make middleware longer for another few lines

site/src/middleware.ts
```diff
@@ -2,10 +2,34 @@ import { Rewrite } from "next/dist/lib/load-custom-routes";
...
+    }
+
+    // Site-Preview is handled in getSiteConfigs since draftMode() is not set yet in middleware (https://github.com/vercel/next.js/issues/52080)
```
nsams, line 19: I don't understand this comment, what line(s) of code does it refer to?

site/src/middleware.ts
```diff
@@ -2,10 +2,34 @@ import { Rewrite } from "next/dist/lib/load-custom-routes";
...
+    const siteConfig = getSiteConfigs().find((siteConfig) => siteConfig.domains.main === host || siteConfig.domains.preliminary === host);
+    if (!siteConfig) {
+        const redirectSiteConfig = getSiteConfigs().find(
+            (siteConfig) =>
+                siteConfig.domains.additional?.includes(host) || (siteConfig.domains.pattern && host.match(new RegExp(siteConfig.domains.pattern))),
+        );
+        if (redirectSiteConfig) {
+            return NextResponse.redirect(redirectSiteConfig.url);
+        }
+        return; // Fallback for Site-Preview
```
nsams, line 31: is this return really correct?

and don't we know the site-preview domain so we can check for it?

site/src/middleware.ts
nsams: this todo should be removed for now

site/src/config.ts
```diff
@@ -1,3 +1,32 @@
...
+        if (preview && preview.scope) {
+            siteConfig = getSiteConfigFromScope(preview.scope as ContentScope);
+        }
+    }
+
+    if (!siteConfig) notFound();
+    return siteConfig;
+}
+
+export function getHost(headers: Headers) {
```
nsams, line 25: maybe rename to `getHostFromHeaders`?

site/src/app/sitemap.ts
```diff
@@ -1,52 +1,61 @@
...
+
 export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
     const sitemap: MetadataRoute.Sitemap = [];
     const graphqlFetch = createGraphQLFetch();
 
-    for (const lang of languages) {
-        const scope = { domain, language: lang };
+    const siteConfigs = getSiteConfigs();
+
+    for (const siteConfig of siteConfigs) {
```
nsams, line 20: a sitemap should only contain sites from one domain (=one siteConfig in this configuration)

site-configs.ts
```diff
@@ -20,16 +24,28 @@ const getSiteConfigs = async (env: Environment): Promise<SiteConfig[]> => {
     return imports.map((imprt, index) => {
         const { domains, ...site } = imprt.default;
 
-        const ret: SiteConfig = {
+        const contentScopes = site.languages.map(
```
nsams, line 27: you could move this into api where `availableContentScopes` is set and omit `contentScopes` from site-config alltogether. What do you think?

site-configs.ts
```diff
@@ -20,16 +24,28 @@ const getSiteConfigs = async (env: Environment): Promise<SiteConfig[]> => {
...
+            (language) => ({
+                domain: site.domain,
+                language,
+            })
+        );
+
+        return {
             ...site,
+            name: `${site.name}`,
```
nsams, line 36:
```suggestion
```

site-configs.ts
```diff
@@ -3,14 +3,18 @@ import { SiteConfig } from "./site-configs.d";
...
+export type Config = Omit<SiteConfig, "domains" | "public"> & Pick<SiteConfig["public"], "domain" | "languages"> & {
     domains: {
         preliminary?: string;
     } & {
         [key in Environment]?: string;
     };
 };
 
+const getUrlFromDomain = (domain: string): string => {
+    return domain.includes("localhost") ? `http://${domain}` : `https://${domain}`;
```
nsams, line 15:
```suggestion
    return domain.endsWith("localhost") ? `http://${domain}` : `https://${domain}`;
```

thomasdax98: Doesn't work because of the port

site-configs.ts
thomasdax98: 0c03a1fa6b1629f214cc5a325625636f287790c8

site/src/app/sitemap.ts
thomasdax98: Oops, I misunderstood your comment https://github.com/vivid-planet/comet-starter/pull/283#discussion_r1677364899

Now: ff43b9e13f2565b11132b62befac4f0fe764116f

site/src/config.ts
thomasdax98: 62aacb1051facc62a31358b242ed3ba9e7d19b56

site/src/middleware.ts
thomasdax98: 12443aabb096641d5abcf0abf33124cf80a5e882

site/src/middleware.ts
thomasdax98: I don't know. It was originally added by @fraxachun and then copied here. Could well be that it's already outdated

I removed it: 4067623e8dac6b9535d0749d60a58ac7320e2fed

site/src/middleware.ts
thomasdax98: 71063d31681863323b89b9cb856cf0a8d4c9b1e5

site-configs.ts
thomasdax98: 2786149c319d0b589d5bf93d01979fe5bb54257b

site-configs.ts
thomasdax98: It could be `startsWith` maybe. But I doubt that we ever have a case where an actual domain contains localhost

```ts
export default {
    name: "Starter Main",
    domain: "main",
    languages: ["en", "de"],
    domains: {
        local: "localhost:3000",
    },
} satisfies Config;
```

site/src/app/IntlProvider.tsx
johnnyomair, line 1: Why was this file moved?

site/src/app/sitemap.ts
```diff
@@ -37,7 +44,7 @@ export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
                         const seoBlock = pageTreeNode.document.seo;
                         if (!seoBlock.noIndex) {
                             sitemap.push({
-                                url: `${process.env.SITE_URL}/${lang}${pageTreeNode.path}`, // TODO support multiple site domains
+                                url: `${getUrlFromDomain(siteConfig.domains.main)}/${scope.language}${pageTreeNode.path}`, // TODO support multiple site domains
```
johnnyomair, line 47:
```suggestion
                                url: `${getUrlFromDomain(siteConfig.domains.main)}/${scope.language}${pageTreeNode.path}`,
```

I believe we can remove this todo

admin/src/config.tsx
```diff
@@ -29,7 +29,7 @@ export function createConfig() {
     };
 }
 
-export type ContentScope = PublicSiteConfig["contentScope"];
+export type ContentScope = SiteConfigContentScope;
```
johnnyomair, line 32: We could also remove this export altogether and always import from ./site-configs

site-configs.ts
johnnyomair: We could also use a Regex

site-configs.ts
```diff
@@ -3,36 +3,40 @@ import { SiteConfig } from "./site-configs.d";
...
+        return {
             ...site,
             domains: {
                 main: domains[env] ?? "",
                 preliminary: env === "prod" ? domains["preliminary"] : undefined,
             },
             preloginEnabled: env === "prod" ? site.preloginEnabled : true,
             public: {
-                contentScope,
+                previewUrl: getUrlFromDomain(domains[env] ?? ""),
```
johnnyomair, line 35: I'm not sure if this is correct. In our setup the preview url is typically a separate deployment, something like <env>-preview.example.com

site/src/app/IntlProvider.tsx
thomasdax98: The [lang] folder was removed in an earlier version and it was never moved back: 7cb3efb4e15848e1e9f4ef4661e601d42beacb50

site-configs.ts
thomasdax98: Done: 0fe7a626ba45ef4eeba0bb0163245e28bd217ca4

site-configs.ts
thomasdax98: I just saw this PR: https://github.com/vivid-planet/comet/pull/2163

So apparently @fraxachun's plan was to not have the previewUrl in the site config at all. So I'll move it to .env

thomasdax98: But then I again have the problem of getting the correct site config in the preview because I can't rely on the domain...

nsams: thing is that the site domain/url used for preview is:

1. not a domain of one of the siteConfigs
2. one domain used for all site configs

so I think we need something not part of the site-config mechanism

site-configs.ts
```diff
@@ -3,36 +3,44 @@ import { SiteConfig } from "./site-configs.d";
...
+export type Config = Omit<SiteConfig, "domains" | "public"> & Pick<SiteConfig["public"], "domain" | "languages"> & {
     domains: {
         preliminary?: string;
     } & {
         [key in Environment]?: string;
     };
 };
 
+const isLocalhost = (domain:string) => {
+    return domain.match(/\b(?:[a-zA-Z0-9-]+\.)?localhost:\d{4}\b/);
```
nsams, line 15: I'm not sure what \b means here, I'd use ^ and $

```suggestion
    return domain.match(/^([a-zA-Z0-9-]+\.)?localhost:\d{4}$/);
```

thomasdax98: 016068ba5900cba61c26aad6a71d14c23c174c35

site-configs.ts
nsams: do you want to merge this first?

thomasdax98: Let's discuss this tomorrow. I don't think the current setup would work correctly in a deployed setting. So we probably won't gain anything from merging it now. Though we could at least continue working on #285 and #286

site/src/middleware.ts
thomasdax98: Apparently, this was more of a TODO and was replaced with https://github.com/vivid-planet/comet-starter/blob/96e28620f5b9b5418297d640341488bbd969b677/site/src/middleware.ts#L40-L49

site-configs/main.ts
```diff
@@ -1,10 +1,16 @@
...
-    languages: ["en", "de"],
-    domains: {
-        local: "localhost:3000",
-    },
-} satisfies Config;
+export default ((env) => {
+    return {
+        name: "Starter Main",
+        domains: {
+            main: "localhost:3000",
```
nsams, line 7: 1. the site config does have a single domain (for a given env), I suppose this must not be an object
2. you should also define production domain (and other envs)
3. It is confusing to have public.domain = "main" vs. domains = "localhost:3000"

thomasdax98: see https://github.com/vivid-planet/comet-starter/pull/316#discussion_r1696468842

site/src/middleware.ts
```diff
@@ -44,8 +86,10 @@ export const config = {
          * - _next/image (image optimization files)
          * - favicon.ico, favicon.svg, favicon.png
          * - manifest.json
+         * and ending with
+         * - .svg (SVG files)
          */
-        "/((?!api|_next/static|_next/image|favicon.ico|favicon.svg|favicon.png|manifest.json).*)",
+        "/((?!api|_next/static|_next/image|favicon.ico|favicon.svg|favicon.png|manifest.json|assets).*)",
```
johnnyomair, line 92: In some of our projects we checked for files in the public folder using a Regex `/\.(.*)$/` (= the file has an extension). Would that be an alternative to the assets/ folder?

admin/src/App.tsx
```diff
@@ -57,13 +58,18 @@ export function App() {
...
                             configs: config.sitesConfig,
                             resolveSiteConfigForScope: (configs, scope: ContentScope) => {
-                                const siteConfig = configs.find((config) => config.contentScope.domain === scope.domain);
+                                const siteConfig = configs.find((config) => {
+                                    return config.domain === scope.domain;
+                                });
+
                                 if (!siteConfig) throw new Error(`siteConfig not found for domain ${scope.domain}`);
+
+                                const previewUrl = process.env.NODE_ENV === "development" ? siteConfig.url : config.previewUrl;
```
johnnyomair, line 67: Should we add a comment for this if?

site-configs.d.ts
```diff
@@ -8,9 +8,10 @@ export type ContentScope = {
...
-        contentScope: ContentScope;
+        domain: string;
+        languages: string[];
     };
 }
 
-export type PrivateSiteConfig = ExtractPrivateSiteConfig<SiteConfig> & { contentScope: ContentScope };
-export type PublicSiteConfig = ExtractPublicSiteConfig<SiteConfig> & { contentScope: ContentScope };
+export type PrivateSiteConfig = ExtractPrivateSiteConfig<SiteConfig> & SiteConfig["public"];
+export type PublicSiteConfig = ExtractPublicSiteConfig<SiteConfig> & SiteConfig["public"];
```
johnnyomair, line 17: This is fixed in https://github.com/vivid-planet/comet/pull/2359, right? Should we wait for this fix in core?

site/public/assets/comet-logo.svg
johnnyomair, line 1: I'm not particularly fond of the assets/ folder...

admin/src/App.tsx
thomasdax98, line 67: The preview URL is now back in .env because there is only one URL in deployed environments.

This does not work locally. We would need a separate domain for this that does not match any SiteConfig (e.g. preview.localhost:3000). Unfortunately, setting the cookies does not work because the redirect from preview.localhost:3000 to localhost:3000 in the API route is a cross-site request and therefore SameSite=Lax cookies are not set.

Instead, we now simply use the site URL from the SiteConfig locally: https://github.com/vivid-planet/comet-starter/pull/283/commits/0b99998b3bd014c6bb514049001fc96c7be11725

This is not quite ideal because the preview behaves differently locally and deployed. But we have not come up with any other good solution.

---

There are now two ways to determine the SiteConfig:

1. Match domain from the URL
2. Read scope parameters in the preview

Deployed we do 1 in “normal operation” and 2 in the preview. Locally we do 1 in normal operation and 1 in the preview. 

We cannot do 2 locally in the preview because the cookies do not work and we therefore do not know that we are in the preview. That's why we have to use different domains locally for the preview.

This has the disadvantage that we never run into variant 2 locally and it is therefore difficult to test.

site-configs.d.ts
thomasdax98: Yes, we can

site/src/app/[domain]/[lang]/[[...path]]/layout.tsx
```diff
@@ -27,7 +25,7 @@ export default async function Layout({ children, params }: PropsWithChildren<{ p
 
             ${headerFragment}
         `,
-        { domain: scope.domain, language: scope.language },
+        { domain: params.domain, language: params.lang },
```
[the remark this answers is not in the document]
Comment (thomasdax98) [reply site/src/app/[domain]/[lang]/[[...path]]/layout.tsx:28]: Yes this works because the URL is rewritten in `createRewriteUrl` also in the preview. So the params are still in the URL

site/src/app/[domain]/sitemap.ts
```diff
@@ -11,8 +12,11 @@ export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
     const sitemap: MetadataRoute.Sitemap = [];
     const graphqlFetch = createGraphQLFetch();
 
-    for (const lang of languages) {
-        const scope = { domain, language: lang };
+    const siteConfig = getSiteConfigForDomain(getHost(headers())); // https://github.com/vercel/next.js/discussions/43179
+
+    for (const language of siteConfig.languages) {
+        const domain = siteConfig.domain;
```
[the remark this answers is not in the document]
Comment (thomasdax98) [reply site/src/app/[domain]/sitemap.ts:18]: I couldn't find a way to access the params in sitemap.ts

site/src/app/[domain]/sitemap.ts
```diff
@@ -11,8 +12,11 @@ export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
     const sitemap: MetadataRoute.Sitemap = [];
     const graphqlFetch = createGraphQLFetch();
 
-    for (const lang of languages) {
-        const scope = { domain, language: lang };
+    const siteConfig = getSiteConfigForDomain(getHost(headers())); // https://github.com/vercel/next.js/discussions/43179
```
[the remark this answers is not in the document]
Comment (thomasdax98) [reply site/src/app/[domain]/sitemap.ts]: I guess I copied it accidentially 163c1fb77ca3edc2b88e143335ea693fcce27107

site/src/middleware.ts
thomasdax98: I tried this in 6aea6a69988be15e46a292c28d2e21eb49ea9e09 but then realised that this also affects sitemap.xml which didn't work anymore afterwards (and probably other special files). So I reverted it again b725c268991de8b1ca2e715e7d1331e54e6c79d5

We could, of course, exclude sitemap.xml explicitly. But the regex may become complex over time

thomasdax98:
> We should also add a `robots.txt` and a custom `not-found.tsx`.

@fraxachun I want to do this in follow-up PRs. There is already a PR for the Not Found Page: https://github.com/vivid-planet/comet-starter/pull/285. I will create one for robots.txt later.

Right now our priority should be to merge this PR as soon as possible with a (relatively) stable site config and a setup for site that works locally. We need to be able to start new projects

site/src/app/[domain]/[lang]/[[...path]]/layout.tsx
```diff
@@ -12,12 +11,11 @@ export const metadata: Metadata = {
     title: "Comet Starter",
 };
 
-export default async function Layout({ children, params }: PropsWithChildren<{ params: { lang: string } }>) {
-    // TODO support multiple domains, get domain by Host header
-    const { scope, previewData } = (await previewParams()) || { scope: { domain, language: params.lang }, previewData: undefined };
-    const graphQLFetch = createGraphQLFetch(previewData);
+export default async function Layout({ children, params }: PropsWithChildren<{ params: { domain: string; lang: string } }>) {
```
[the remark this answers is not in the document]
Comment (thomasdax98) [reply site/src/app/[domain]/[lang]/[[...path]]/layout.tsx]: b5bad9566d77cf8cbd4a58fb2605d7057ddc8410

site/src/middleware.ts
nsams: > the file has an extension

that might be the case for so many cases, I think a solution where we can detect assets is needed.

site/src/app/[domain]/sitemap.ts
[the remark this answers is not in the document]
Comment (johnnyomair) [reply site/src/app/[domain]/sitemap.ts:18]: Okay, maybe we could re-add the link to the discussion and add a comment why it's needed: https://github.com/vivid-planet/comet-starter/pull/283#discussion_r1696631429

site/src/middleware.ts
johnnyomair: assets/ is okay for now, we can still change this in the future.

site-configs.d.ts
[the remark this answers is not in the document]
thomasdax98: cc0a044a0a7038728fb635c6fb33069770897ae2

site/src/app/[domain]/sitemap.ts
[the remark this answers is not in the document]
Comment (thomasdax98) [reply site/src/app/[domain]/sitemap.ts:20]: Done in 96173cedc5e6dc3710203514004318dd5de796aa

johnnyomair (dismissed):

johnnyomair (dismissed):

johnnyomair (dismissed):

site/src/middleware.ts
```diff
@@ -1,12 +1,54 @@
...
+        if (redirectSiteConfig) {
+            return NextResponse.redirect(redirectSiteConfig.url);
+        }
+
+        // Site-Preview
+        const sitePreviewParams = await previewParams({ skipDraftModeCheck: true });
+
+        if (sitePreviewParams?.scope) {
+            const siteConfig = getSiteConfigForScope(sitePreviewParams.scope as ContentScope);
+            headers.set("x-forwarded-host", siteConfig.domains.main);
```
nsams, line 45: I think that is wrong (domains.main) and not needed

site/src/config.ts
```diff
@@ -1,3 +1,36 @@
...
+export function getSiteConfigs() {
+    if (!siteConfigs) {
+        const json = typeof window === "undefined" ? process.env.NEXT_PUBLIC_SITE_CONFIGS : env("NEXT_PUBLIC_SITE_CONFIGS");
+        if (!json) throw new Error("process.env.NEXT_PUBLIC_SITE_CONFIGS must be set.");
+        siteConfigs = JSON.parse(json) as SiteConfig[];
+    }
+    return siteConfigs;
+}
+
+export function getHost(headers?: Headers) {
```
nsams, line 30: headers should not be options, it's never called without headers. And maybe call it getHostFromHeaders, and don't place it into config - it is unrelated to config.

site/src/middleware.ts
```diff
@@ -1,12 +1,54 @@
...
+        return NextResponse.next({ request: { headers } });
+    }
+
+    // Site-Preview
+    const sitePreviewParams = await previewParams({ skipDraftModeCheck: true });
+
+    if (sitePreviewParams?.scope) {
+        const siteConfig = getSiteConfigForScope(sitePreviewParams.scope as ContentScope);
+        headers.set("x-forwarded-host", siteConfig.domains.main);
+        return NextResponse.rewrite(createRewriteUrl(request, siteConfig.domain), { request: { headers } });
```
nsams, line 35: you could directly access `sitePreviewParams.scope.domain`, no need to access the whole siteConfig.

getSiteConfigForScope can be removed then

nsams, line 35: If you return early here, you loose everything from here below of the middleware for site preview. `/dam/` rewrite, redirects and rewrites. I think we need all of those in site preview as well.

admin/src/App.tsx
```diff
@@ -57,8 +58,15 @@ export function App() {
...
-                                const siteConfig = configs.find((config) => config.contentScope.domain === scope.domain);
+                                const siteConfig = configs.find((config) => {
+                                    return config.domain === scope.domain;
+                                });
+
                                 if (!siteConfig) throw new Error(`siteConfig not found for domain ${scope.domain}`);
+
+                                // In a deployed setting, we use one preview URL for all scopes.
+                                // Locally, we use the site URL for each scope because setting the necessary cookies for the preview doesn't work on localhost.
+                                // see https://github.com/vivid-planet/comet-starter/pull/283/files#r1696858091 for more infos
```
nsams, line 69:
```suggestion
```

this comment is outdated

nsams (approved):

site/src/middleware.ts
```diff
@@ -1,12 +1,59 @@
...
+    const sitePreviewParams = await previewParams({ skipDraftModeCheck: true });
+    if (sitePreviewParams?.scope) {
+        const siteConfig = getSiteConfigs().find((siteConfig) => siteConfig.domain === sitePreviewParams.scope.domain);
+        if (siteConfig) return siteConfig;
+    }
+    return getSiteConfigs().find((siteConfig) => siteConfig.domains.main === host || siteConfig.domains.preliminary === host);
+}
+
+// Used for getting SiteConfig in server-components where params is not available (e.g. sitemap, not-found - see https://github.com/vercel/next.js/discussions/43179)
+export async function getSiteConfig() {
```
nsams, line 26: I'm not sure middleware is the right place for this function

nsams: I understand.

You could put it into it's own file?

thomasdax98 (approved):

PR #284, opened 2024-07-04
[RFC] Validate scope in GraphQL schema

johnnyomair:
## Motivation

In recent pen tests I've noticed that we don't validate the value of the provide scope dimensions. For instance, one attempt during a pen test failed with a DB driver exception because the provided scope domain was: 

```
a:2:{i:7;O:15:"Zend\Log\Logger":1:{s:10:"*writers";a:1:{i:0;O:20:"Zend\Log\Writer\Mail":3:{s:15:"*eventsToMail";a:1:{i:0;i:0;}s:21:"*subjectPrependText";s:0:"";s:24:"*numEntriesPerPriority";a:1:{i:0;O:14:"Zend\Tag\Cloud":2:{s:7:"*tags";a:1:{i:0;s:0:"";}s:15:"*tagDecorator";O:34:"Zend\Tag\Cloud\Decorator\HtmlCloud":3:{s:12:"*separator";s:0:"";s:10:"*escaper";O:20:"Zend\Escaper\Escaper":1:{s:18:"*htmlAttrMatcher";a:2:{i:0;O:23:"Zend\Filter\FilterChain":1:{s:10:"*filters";O:13:"SplFixedArray":2:{i:0;a:2:{i:0;O:14:"Zend\Json\Expr":1:{s:13:"*expression";s:11:"tenable.was";}i:1;s:10:"__toString";}i:1;s:13:"gethostbyname";}}i:1;s:6:"filter";}}s:11:"*htmlTags";a:1:{s:1:"h";a:1:{s:1:"a";s:1:"!";}}}}}}}}i:7;i:7;}
```

This could be prevented on the GraphQL layer by using more strict types for the content scope, for instance, enums.

## Example

Before: Anything is allowed, API responds with no results (best case)

before

After: Only specific values are allowed, API fails with a validation error (no leaks due to disabling field suggestions)

after

## Alternatives

* Validate using class-validator (only possible by using args classes)
* Ignore

What do you think?

### DIFF
--- a/admin/src/common/ContentScopeProvider.tsx
+++ b/admin/src/common/ContentScopeProvider.tsx
@@ -13,12 +13,11 @@ import {
     useSitesConfig,
 } from "@comet/cms-admin";
 import { SitesConfig } from "@src/config";
+import { GQLDomain, GQLLanguage } from "@src/graphql.generated";
 
-type Domain = "main" | "secondary" | string;
-type Language = "en" | string;
 export interface ContentScope {
-    domain: Domain;
-    language: Language;
+    domain: GQLDomain;
+    language: GQLLanguage;
 }
 
 // convenince wrapper for app (Bind Generic)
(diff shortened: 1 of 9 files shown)


dkarnutsch (dismissed):
Nice 👍

nsams (approved):

max-debug022 (approved):

dkarnutsch (approved):

johnnyomair:
I'd like to wait on @fraxachun's opinion for this.

thomasdax98 (approved):

nsams:
I wonder how well this plays together with #283

johnnyomair:
> I wonder how well this plays together with #283

It would be possible to hard code all available combinations, or create the scope DTOs dynamically based on the site config (overkill?).

johnnyomair:
We discussed this in today's meeting and decided to not add it for the following reasons:

1. The content scopes would need to be duplicated between site configs and the scope DTOs (could be a lot)
2. The user permissions module should deny access for all non-existing scopes (except public API handlers, but those should be replace with BFFs)

PR #285, opened 2024-07-05
Add not found page

magdalenaxm:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-806

<details>

Screenshot 2024-08-06 at 07 04 00

</details>


johnnyomair:

site/src/app/[domain]/[language]/layout.tsx
```diff
@@ -1,5 +1,12 @@
+import { gql, previewParams } from "@comet/cms-site";
 import { IntlProvider } from "@src/app/[domain]/[language]/IntlProvider";
+import { GQLLayoutQuery, GQLLayoutQueryVariables } from "@src/app/[domain]/[language]/layout.generated";
```
Comment (nsams) [inline site/src/app/[domain]/[language]/layout.tsx:3]: please use relative import

nsams (approved):

site/src/app/[domain]/[language]/not-found-content.tsx
```diff
@@ -0,0 +1,22 @@
+"use client";
+
+import Link from "next/link";
+import { useParams } from "next/navigation";
+import React from "react";
+import { FormattedMessage } from "react-intl";
+
+export const NotFoundContent = (): React.ReactElement => {
```
Comment (johnnyomair) [inline site/src/app/[domain]/[language]/not-found-content.tsx:8]:
```suggestion
export function NotFoundContent() {
```

Nit: I'd use the same style for function components everywhere

site/src/app/[domain]/[language]/not-found-content.tsx
```diff
@@ -0,0 +1,22 @@
...
+    const { language } = useParams();
+
+    return (
+        <>
+            <h1>
+                <FormattedMessage id="notFound.headline" defaultMessage="Oops. Something went wrong." />
```
Comment (johnnyomair) [inline site/src/app/[domain]/[language]/not-found-content.tsx:14]: Do we really want such a goofy not found message per default? 😁

site/src/app/[domain]/[language]/not-found-content.tsx
```diff
@@ -0,0 +1,22 @@
...
+            </h1>
+            <FormattedMessage id="notFound.text" defaultMessage="Sorry, the selected page is currently not available." />
+            <Link href={`/${language}`}>
+                <FormattedMessage id="notFound.link" defaultMessage="Back to Homepage" />
```
Comment (johnnyomair) [inline site/src/app/[domain]/[language]/not-found-content.tsx:18]:
```suggestion
                <FormattedMessage id="notFound.link" defaultMessage="Back to Home" />
```

site/src/app/[domain]/[language]/not-found-content.tsx
Comment (thomasdax98) [reply site/src/app/[domain]/[language]/not-found-content.tsx]: I changed the text 94f54fe4da244084c1c1924084779fefaa22705e

Bildschirmfoto 2024-08-06 um 09 42 06

site/src/app/[domain]/[language]/not-found-content.tsx
Comment (thomasdax98) [reply site/src/app/[domain]/[language]/not-found-content.tsx]: 94f54fe4da244084c1c1924084779fefaa22705e

site/src/app/[domain]/[language]/not-found-content.tsx
Comment (thomasdax98) [reply site/src/app/[domain]/[language]/not-found-content.tsx]: 50a8b59fed07b2b5d206057f088fc5b48049ce03

site/src/app/[domain]/[language]/layout.tsx
Comment (thomasdax98) [reply site/src/app/[domain]/[language]/layout.tsx]: af04bcc81539ea1ef4d009eb98da281ba3e3f2b4

site/src/app/[domain]/[language]/not-found-content.tsx
```diff
@@ -0,0 +1,33 @@
...
+                <FormattedMessage id="notFound.headline" defaultMessage="The requested page does not exist" />
+            </Typography>
+            <Typography variant="p200">
+                <FormattedMessage
+                    id="notFound.text"
+                    defaultMessage="We apologize, but the page you are looking for cannot be found. Please try navigating to a different page or go back to the homepage. Thank you."
+                />
+            </Typography>
+            <Typography variant="p200">
+                <Link href={`/${language}`}>
```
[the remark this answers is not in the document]
Comment (thomasdax98) [reply site/src/app/[domain]/[language]/not-found-content.tsx:27]: The layout uses the not found page of the layer above, so it falls back to the Next.js default:

Bildschirmfoto 2024-08-07 um 11 20 55

We could also add an extra not found page for that. The problem is, we can't use FormattedMessage there because we have no IntlProvider since we don't know which language we are in. 

Right now, we only have a not found page within the language scope. We could also add one for fallback which contains hardcoded English

site/src/app/[domain]/[language]/layout.tsx
```diff
@@ -1,19 +1,48 @@
...
+import { createGraphQLFetch } from "@src/util/graphQLClient";
+import { notFound } from "next/navigation";
+
+import { GQLLayoutQuery, GQLLayoutQueryVariables } from "./layout.generated";
+
+export default async function Page({
+    children,
+    params: { domain, language },
+}: {
+    children: React.ReactNode;
```
Comment (johnnyomair) [inline site/src/app/[domain]/[language]/layout.tsx:16]: React isn't imported in this file 🤔 Also, @SebiVPS is trying to get rid of the React barrel imports. You can use PropsWithChildren here

johnnyomair:
@thomasdax98 what's the status here?

thomasdax98:
@johnnyomair I simplified it once more. Now it's ready from my perspective

site/src/layout/header/ServerSideHeader.tsx
johnnyomair, line 1: It's odd that we need an additional component to fetch the header data for the non-internationalized not found page. Wouldn't it be better if the not found page fetched the header and passes it to the component, as we do in the layout?

thomasdax98: If I do that, I need to make the query in the not found page and in the layout. I wanted to avoid the code duplication, so I moved it to this shared component. I can't put it directly in the `Header` since it's a client component.

Would you prefer the code duplication?

johnnyomair: > Would you prefer the code duplication?

I'm not sure. I'd prefer to have the solution that's the most easy to understand. It took me a while to understand your change.

Would it be possible to reuse the layout from /[domain]/[language]/layout.tsx in /[domain]/layout.tsx with a default language?

nsams: Moving (almost everything) from `layout` to `Header` might not be ideal: in real world applications the layout contains probably more than just a header, for example a footer.

PR #282, opened 2024-07-01
Add footer

jomunker:
---

This adds a basic footer, which can be configured in the admin. The footer customization can be found under `Project Snips/Footer`.

## PR Checklist

-   [x] Link to the respective task if one exists: COM-872
-   [x] Provide screenshots/screencasts if the change contains visual changes

<details>
    <summary>Screenshots/screencasts</summary>
    

## Screencast Responsive
https://github.com/user-attachments/assets/4f997975-af29-41f7-88df-a8b0f2505a1c

## Admin
image

</details>

PR #288, opened 2024-07-08
Add `BillboardTeaserBlock`

SebiVPS:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-838
-   [x] Add overlay

<details>
Screenshot 2024-07-18 at 16 18 34

</details>


site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,115 @@
...
+    flex: 1;
+    flex-direction: column;
+    justify-content: center;
+    align-items: ${({ $alignItems }) => $alignItems};
+`;
+
+const BaseImage = styled.div`
+    img {
+        object-fit: cover;
+    }
```
jamesricky, line 76: This should probably be handled by the Image-/Media-Block itself.

SebiVPS: yeah, need to check that with the updated image/media component

nsams:
shouldn't the content be aligned just like other content? (becasue you don't use the page grid)

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+    ),
+    { label: "Billboard Teaser" },
+);
+
+const Root = styled.div`
+    position: relative;
+    overflow: hidden;
+`;
+
+const AbsoluteGridRoot = styled(PageLayout)`
```
SebiVPS, line 48: @nsams it is used here

jamesricky (approved):

johnnyomair:

jamesricky (dismissed):

admin/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,37 @@
...
+        displayName: <FormattedMessage id="billboardTeaserBlock.displayName" defaultMessage="Billboard Teaser" />,
+        blocks: {
+            media: {
+                block: MediaBlock,
+                title: <FormattedMessage id="billboardTeaserBlock.media" defaultMessage="Media" />,
+                hiddenInSubroute: true,
+            },
+            heading: {
+                block: HeadingBlock,
+            },
```
johnnyomair, line 21: This probably needs a hiddenInSubroute as well

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+                <MediaBlock data={media} aspectRatio="16x9" />
+            </ImageDesktop>
+            <ImageLargeDesktop>
+                <MediaBlock data={media} aspectRatio="3x1" />
+            </ImageLargeDesktop>
+            <AbsoluteGridRoot grid>
+                <Content $alignItems={alignmentMap[heading.textAlignment]}>
+                    <HeadingBlock data={heading} />
+                    <RichTextBlock data={text} />
+                    {callToActionList.blocks.length > 0 && <CallToActionListBlock data={callToActionList} />}
```
johnnyomair, line 35: Shouldn't this length check be inside CallToActionListBlock?

admin/src/documents/pages/blocks/BillboardTeaserBlock.tsx
SebiVPS: I do not know exactly why, but it does not need ahiddenInSubroute

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
SebiVPS: removed it

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -1,94 +1,60 @@
...
+    SUB: (children, { key }) => <sub key={key}>{children}</sub>,
+    SUP: (children, { key }) => <sup key={key}>{children}</sup>,
+    STRIKETHROUGH: (children, { key }) => <s key={key}>{children}</s>,
+};
+
 /**
  * Define the renderers
  */
-export const defaultRichTextRenderers: Renderers = {
+const getDefaultRenderers = (colorInverted?: boolean): Renderers => ({
```
johnnyomair, line 33: In #282 @jomunker removed the default text color from the rich text block, which is a good idea. Then the inverted color could be set in the billboard teaser block.

site/src/common/components/Typography.tsx
```diff
@@ -195,23 +195,31 @@ export interface TypographyProps extends HTMLAttributes<HTMLElement> {
...
+        {...restProps}
+    >
         {children}
     </Text>
 );
 
 interface TextProps {
     $variant: TypographyVariant;
     $bottomSpacing?: boolean;
+    $colorInverted?: boolean;
```
johnnyomair, line 217: Typography shouldn't set a color at all, then this could be done in the billboard teaser block

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,106 @@
...
+    padding: ${({ theme }) => theme.spacing.D200} 0;
+    display: flex;
+    flex: 1;
+    flex-direction: column;
+    justify-content: center;
+    align-items: ${({ $alignItems }) => $alignItems};
+`;
+
+const BaseImage = styled.div<{ $backgroundOpacity: string }>`
+    filter: brightness(calc(100% - ${({ $backgroundOpacity }) => $backgroundOpacity}%));
```
johnnyomair, line 67: You could also set a transparent background on the AbsoluteGridRoot component. I don't know what's better though 🤷🏼‍♂️

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { MediaBlock } from "@src/common/blocks/MediaBlock";
+import { RichTextBlock } from "@src/common/blocks/RichTextBlock";
+import { PageLayout } from "@src/layout/PageLayout";
+import { CSSProperties } from "react";
+import styled from "styled-components";
+
+const alignmentMap: Record<HeadingBlockData["textAlignment"], CSSProperties["alignItems"]> = {
+    Left: "left",
+    Center: "center",
+};
```
johnnyomair, line 14: Using the headline alignment to align everything seems arbitrary to me... Did you discuss this with UX?

api/src/documents/pages/blocks/billboard-teaser.block.ts
```diff
@@ -0,0 +1,70 @@
...
+    percentage90 = "90",
+    percentage80 = "80",
+    percentage70 = "70",
+    percentage60 = "60",
+    percentage50 = "50",
+    percentage40 = "40",
+    percentage30 = "30",
+    percentage20 = "20",
+    percentage10 = "10",
+}
```
johnnyomair, line 29: Instead of an enum you could also use number or string 🤔

mariokemetinger (approved):

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
SebiVPS: For me it seems superfluous to have a second  alignment select with same options. But if thats what we want i'll add that.

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
SebiVPS: I do not know what's better either ... @jamesricky @mariokemetinger ?

api/src/documents/pages/blocks/billboard-teaser.block.ts
SebiVPS: i think number is not supported by select field in admin, ill make it a string then 👍

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+                </Content>
+            </AbsoluteGridRoot>
+        </Root>
+    ),
+    { label: "Billboard Teaser" },
+);
+
+const Root = styled(PageLayout)`
+    position: relative;
+    overflow: hidden;
```
jamesricky, line 40: I don't think this should be needed 🤔

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
jamesricky: I'd just add a pseudo-element that overlays the image with a black `background-color` what has a variable `opacity` value.

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
SebiVPS: if someone adds a lot of content it will overflow, but thats content ...

SebiVPS:
> I'd just add a pseudo-element that overlays the image with a black background-color what has a variable opacity value.

i found the one liner quite elegant

site/src/documents/pages/blocks/BillboardTeaserBlock.tsx
SebiVPS: removed that, according to the design the content should always be centered

jamesricky (dismissed):

site/src/common/blocks/HeadingBlock.tsx
```diff
@@ -7,51 +7,21 @@ import { CSSProperties } from "react";
 import { Renderers } from "redraft";
 import styled from "styled-components";
 
-import { defaultRichTextRenderers, RichTextBlock } from "./RichTextBlock";
+import { createTextBlockRenderFn, defaultRichTextInlineStyleMap, RichTextBlock } from "./RichTextBlock";
 
 const eyebrowRenderers: Renderers = {
-    inline: defaultRichTextRenderers.inline,
+    inline: defaultRichTextInlineStyleMap,
```
johnnyomair, line 13: Why did you change this?

api/src/documents/pages/blocks/billboard-teaser.block.ts
johnnyomair: Number is supported, and better for validation: 2087247d8464318a4617782be3971f9593401d05

admin/src/documents/pages/blocks/BillboardTeaserBlock.tsx
```diff
@@ -0,0 +1,60 @@
...
+    { value: 100, label: "100%" },
+    { value: 90, label: "90%" },
+    { value: 80, label: "80%" },
+    { value: 70, label: "70%" },
+    { value: 60, label: "60%" },
+    { value: 50, label: "50%" },
+    { value: 40, label: "40%" },
+    { value: 30, label: "30%" },
+    { value: 20, label: "20%" },
+    { value: 10, label: "10%" },
```
thomasdax98, line 20: 0 - 90 instead of 10 - 100

thomasdax98: 741189f6f99e8eb1ce0afdda4c651f6bda15c455

site/src/common/blocks/HeadingBlock.tsx
thomasdax98: Unfortunately, I can't comment on that. So we have to wait for @SebiVPS

SebiVPS: This changes has been made during necessary code refactoring when adding the `colorInverted` prop. The prop was then removed again when the text color was added to the gobal styles.

These changes do not need to be part of this PR. I think the changes make the code a little bit more readable and usable. As they are indeed no part of this PR you can remove them and, if you want, make a new PR with these adaptions?

johnnyomair:
I'm not sure if the background opacity makes sense from a UX perspective:

If I change it to 10%, the background is very visible:

background opacity 10

If I change it to 90%, the background is barely visible:

background opacity 90

Shouldn't this be the other way around?

thomasdax98:
@SebiVPS Please rename "background opacity" to "overlay". 

"background opacity" is unintuitive because 0 means full opacity and 90 means almost no opacity.

jamesricky (approved):

site/src/common/blocks/HeadingBlock.tsx
johnnyomair: We can keep it in this PR

johnnyomair (approved):

thomasdax98 (approved):

mariokemetinger (approved):

PR #292, opened 2024-07-11
RichTextBlock: only render `div` when `disableLastBottomSpacing` is set

SebiVPS:
prevents invalid html when div is rendered inside p tag when a Typgraphy is wrapped around the RichTextBlock

### DIFF
--- a/site/src/common/blocks/RichTextBlock.tsx
+++ b/site/src/common/blocks/RichTextBlock.tsx
@@ -145,7 +145,7 @@ export const PageContentRichTextBlock = (props: RichTextBlockProps) => (
     </PageLayout>
 );
 
-const NoLastBottomSpace = styled.div`
+const DisableLastBottomSpacing = styled.div`
     ${({ theme }) =>
         css`
             > *:last-child {
--- a/site/src/common/blocks/RichTextBlock.tsx
+++ b/site/src/common/blocks/RichTextBlock.tsx
@@ -130,7 +130,7 @@ export const RichTextBlock = withPreview(
 
         return (
             <PreviewSkeleton title="RichText" type="rows" hasContent={hasRichTextBlockContent(data)}>
-                {disableLastBottomSpacing ? <NoLastBottomSpace>{rendered}</NoLastBottomSpace> : rendered}
+                {disableLastBottomSpacing ? <DisableLastBottomSpacing>{rendered}</DisableLastBottomSpacing> : rendered}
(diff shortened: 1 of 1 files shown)


site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -145,7 +145,9 @@ export const PageContentRichTextBlock = (props: RichTextBlockProps) => (
     </PageLayout>
 );
 
-const Root = styled.div<{ $disableLastBottomSpacing?: boolean }>`
+const Root = styled.span<{ $disableLastBottomSpacing?: boolean }>`
+    display: block;
```
johnnyomair, line 149: I'm not sure if this is the way to go: Why do you need `display: block`? Using an inline tag with block styling to workaround the React warning feels weird...

SebiVPS: updated like discussed 👍

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -130,7 +130,7 @@ export const RichTextBlock = withPreview(
 
         return (
             <PreviewSkeleton title="RichText" type="rows" hasContent={hasRichTextBlockContent(data)}>
-                <Root $disableLastBottomSpacing={disableLastBottomSpacing}>{rendered}</Root>
+                {disableLastBottomSpacing ? <NoLastBottomSpace>{rendered}</NoLastBottomSpace> : rendered}
```
johnnyomair, line 133:
```suggestion
                {disableLastBottomSpacing ? <DisableLastBottomSpacing>{rendered}</DisableLastBottomSpacing> : rendered}
```

site/src/common/blocks/RichTextBlock.tsx
```diff
@@ -145,9 +145,8 @@ export const PageContentRichTextBlock = (props: RichTextBlockProps) => (
     </PageLayout>
 );
 
-const Root = styled.div<{ $disableLastBottomSpacing?: boolean }>`
-    ${({ theme, $disableLastBottomSpacing }) =>
-        $disableLastBottomSpacing &&
+const NoLastBottomSpace = styled.div`
```
johnnyomair, line 148:
```suggestion
const DisableLastBottomSpacing = styled.div`
```

jamesricky (approved):

johnnyomair (approved):

PR #293, opened 2024-07-11
Update Comet v7.0.0-beta.4

SebiVPS:


site/next.config.js
```diff
@@ -61,6 +61,7 @@ const nextConfig = {
                                 frame-ancestors ${process.env.ADMIN_URL};
                                 upgrade-insecure-requests; 
                                 block-all-mixed-content;
+                                frame-src youtube.com https://www.youtube.com https://www.youtube-nocookie.com/;
```
SebiVPS, line 64: @johnnyomair without this, the YouTube video does not play

johnnyomair: @max-debug022 this should be okay, right?

max-debug022: Yes. Could be simplified to `frame-src 'self' https://*.youtube.com https://www.youtube-nocookie.com/;` though.

johnnyomair:
```suggestion
                                frame-src 'self' https://*.youtube.com https://*.youtube-nocookie.com;
```

johnnyomair (approved):

PR #295, opened 2024-07-11
Prevent schema.gql write in production

johnnyomair:
Some of our customers use read-only file systems in deployed environments. We therefore can't write the schema.gql file. Disabling it for production shouldn't be a problem, since our only use case (for now) is generating the types for Site and Admin.


dkarnutsch (approved):
I would add a comment here describing this line

thomasdax98 (dismissed):

thomasdax98 (approved):

dkarnutsch (approved):

PR #304, opened 2024-07-16
Remove legacy behavior for link blocks

johnnyomair:
Since Next 13 the `Link` component no longer requires a child `<a>` tag by default. This new behavior is now used by our `InternalLinkBlock`, which uses `Link` internally. To ensure that all other link blocks (e.g., `ExternalLinkBlock`) can be styled in the application, we now render an `<a>` tag for all those blocks.

This change adds the `className` prop to the `LinkBlock` to allow styling the block. All link block usages have been adapted to directly style the block.

---

COM-936

### DIFF
--- a/site/src/common/components/Button.tsx
+++ b/site/src/common/components/Button.tsx
@@ -18,7 +18,7 @@ export const Button = styled.button<{ variant?: ButtonVariant }>`
     font-weight: 700;
     line-height: 110%;
 
-    ${({ variant = "contained", disabled }) =>
+    ${({ variant = "contained" }) =>
         css`
             ${buttonVariantStyle[variant]}
 
@@ -26,13 +26,6 @@ export const Button = styled.button<{ variant?: ButtonVariant }>`
                 pointer-events: none;
                 ${disabledButtonVariantStyle[variant]};
             }
-
-            /* <a> doesn't support :disabled selector */
-            ${disabled &&
-            css`
-                pointer-events: none;
-                ${disabledButtonVariantStyle[variant]};
-            `}
         `};
 `;
 
--- a/site/src/common/components/Button.tsx
+++ b/site/src/common/components/Button.tsx
@@ -1,26 +1,40 @@
-import { forwardRef, HTMLAttributes, RefObject } from "react";
 import styled, { css } from "styled-components";
 
 export type ButtonVariant = "contained" | "outlined" | "text";
 
-type ButtonProps = {
-    variant?: ButtonVariant;
-    disabled?: boolean;
-} & (HTMLAttributes<HTMLButtonElement> | (HTMLAttributes<HTMLAnchorElement> & Pick<HTMLAnchorElement, "href" | "target">));
+export const Button = styled.button<{ variant?: ButtonVariant }>`
+    display: inline-flex;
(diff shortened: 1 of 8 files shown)


site/src/common/components/Button.tsx
```diff
@@ -1,26 +1,40 @@
...
+                pointer-events: none;
+                ${disabledButtonVariantStyle[variant]};
+            }
+
+            /* <a> doesn't support :disabled selector */
+            ${disabled &&
+            css`
+                pointer-events: none;
+                ${disabledButtonVariantStyle[variant]};
+            `}
```
johnnyomair, line 35: We could discuss removing this altogether. Disabled links don't make any sense to me... In the rich text block we don't render the link, and in link lists we hide the link.

johnnyomair: Also, this isn't enough for accessibility. We would need something like `<a role="link" aria-disabled="true">Learn something!</a>`

But this isn't recommended:

> It is NOT RECOMMENDED to use aria-disabled="true" on an a element with an href attribute.

https://w3c.github.io/html-aria/#el-a

But it works with `role="link"` 😂 

> Where it is [not recommended](https://w3c.github.io/html-aria/#el-a) to use the aria-disabled attribute on an <a href> element, it’s perfectly fine to use this attribute on a role=link.

https://www.scottohara.me/blog/2021/05/28/disabled-links.html

In conclusion, I'd still remove it 😁

jamesricky: I agree, probably makes sense to remove it 😄

site/src/layout/header/Header.tsx
```diff
@@ -68,13 +72,17 @@ const TopLevelLinkContainer = styled.li`
...
     text-decoration: none;
     padding: 5px 10px;
 
     &:hover {
         text-decoration: underline;
     }
+
+    &.active {
+        color: ${({ theme }) => theme.palette.primary.main};
+    }
```
jamesricky, line 85: The `active` class-name could easily be confused with the `a:active` selector. 
Maybe change it to something else... `selected`?

nsams (approved):

SebiVPS (approved):

site/src/layout/header/Header.tsx
johnnyomair: This was inspired by React Router, which also uses active: https://v5.reactrouter.com/web/api/NavLink/activeclassname-string

I believe "active" is the term that most devs use for the currently active page, should we really change that?

site/src/common/components/Button.tsx
johnnyomair: Removed: c1d8bdf9ee123c0d660a8559338aa5607ecfb2b2

site/src/layout/header/Header.tsx
jamesricky: For me at least, "active" on `<a />` means "currently being clicked", not "page I'm currently on". 
But if react router also uses that name it's probably fine.

jamesricky (approved):

thomasdax98 (approved):

mariokemetinger (approved):

PR #307, opened 2024-07-17
Throw error for duplicate messages with different default messages

johnnyomair:


admin/package.json
```diff
@@ -10,13 +10,13 @@
         "gql:types": "graphql-codegen",
         "gql:watch": "graphql-codegen --watch",
         "prelint": "npm run intl:compile && run-p gql:types generate-block-types",
-        "lint": "run-p lint:eslint lint:tsc",
+        "lint": "run-p lint:eslint lint:tsc intl:extract",
```
johnnyomair, line 13: Added it to lint so it also fails for the lint workflow in PRs (message extraction is only in main)

dkarnutsch (approved):

thomasdax98 (approved):

nsams (approved):

PR #309, opened 2024-07-19
Disable last bottom spacing for `RichTextBlock` in page content

SebiVPS:
We decided to remove the margins top/bottom from the Page Content Blocks. We try the "do it yourself" way and give to content creators freedom wich spacing they want to use.


jamesricky:
Didn't we want the `RichTextBlock` to be an exception? 
Because you would want the normal paragraph-spacing between this block and e.g. a Button 🤔

SebiVPS:
> Didn't we want the `RichTextBlock` to be an exception? Because you would want the normal paragraph-spacing between this block and e.g. a Button 🤔

It would not be consistent this way, i think. What space would be after the button, when after the button another `RichTextBlock` would be placed there is also no paragraph-spacing.

jamesricky (approved):

johnnyomair (approved):
If we add it here as well,

mariokemetinger (approved):

PR #315, opened 2024-07-29
Add scope to redirects

jomunker:
---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-878


nsams (dismissed):

api/src/db/migrations/Migration20240729171532.ts
```diff
@@ -0,0 +1,7 @@
+import { Migration } from "@mikro-orm/migrations";
+
+export class Migration20240729171532 extends Migration {
+    async up(): Promise<void> {
+        this.addSql('alter table "Redirect" add column "scope_domain" text not null;');
```
johnnyomair, line 5: You'd need to drop all redirects before adding this non-nullable column. But wouldn't it be better to add it to the one migration we already have?

api/src/documents/pages/pages.module.ts
```diff
@@ -1,11 +1,12 @@
+import { RedirectsModule } from "@comet/cms-api";
 import { MikroOrmModule } from "@mikro-orm/nestjs";
 import { Module } from "@nestjs/common";
 
 import { Page } from "./entities/page.entity";
 import { PagesResolver } from "./pages.resolver";
 
 @Module({
-    imports: [MikroOrmModule.forFeature([Page])],
+    imports: [MikroOrmModule.forFeature([Page]), RedirectsModule],
```
johnnyomair, line 9: I don't think this import is needed

site/src/middleware.ts
```diff
@@ -1,25 +1,29 @@
+import { domain } from "@src/config";
```
johnnyomair, line 1: Please merge main and get the domain from the site config now that #283 has been merged

site/src/redirects/redirects.ts
```diff
@@ -113,9 +115,12 @@ const createApiRedirects = async (): Promise<Map<string, Redirect>> => {
 
 type Redirect = { destination: string; permanent: boolean; has?: RouteHas[] | undefined };
 
-export const createRedirects = async () => {
-    const key = `redirects`;
+export const createRedirects = async (scope: GQLRedirectScope) => {
+    const key = `redirects-${JSON.stringify(scope)}`;
```
johnnyomair, line 119:
```suggestion
    const key = `redirects-${scope.domain}`;
```

api/src/db/migrations/Migration20240729171532.ts
jomunker: Why? It doesn't remove sth but just adds a column. Sure but that wouldn't make a difference for the first point right?

api/src/documents/pages/pages.module.ts
johnnyomair: Remove it and see if the API still starts

api/src/db/migrations/Migration20240729171532.ts
johnnyomair: > Sure but that wouldn't make a difference for the first point right?

You're right.

> But wouldn't it be better to add it to the one migration we already have?

How ever, I'd still at this line to the initial migration, which should be fine for all new projects.

api/src/documents/pages/pages.module.ts
jomunker: https://github.com/vivid-planet/comet-starter/pull/315/commits/712c7d6bcdfd9d2ae9dcbd73c86dd400f52ae9a1

api/src/db/migrations/Migration20240729171532.ts
jomunker: https://github.com/vivid-planet/comet-starter/pull/315/commits/e9a0bb4ae89507c158ee9f6bcdb7d9f2afcb9252

site/src/middleware.ts
```diff
@@ -41,6 +42,7 @@ export async function middleware(request: NextRequest) {
     }
 
     const siteConfig = await getSiteConfigForHost(host);
+    const scope = siteConfig?.scope.domain ? { domain: siteConfig?.scope.domain } : undefined;
```
johnnyomair, line 45: You could get the scope after the if below, then you shouldn't need the undefined handling I suppose.

api/src/db/migrations/Migration20220721122758.ts
```diff
@@ -19,5 +19,7 @@ export class Migration20220721122758 extends Migration {
             'alter table "PageTreeNode" add constraint "PageTreeNode_parentId_foreign" foreign key ("parentId") references "PageTreeNode" ("id") on update cascade on delete set null;',
         );
         this.addSql('create index "PageTreeNode_parentId_index" on "PageTreeNode" ("parentId");');
+
+        this.addSql('alter table "Redirect" add column "scope_domain" text not null;');
```
dkarnutsch, line 23: I don't think it is good practice to alter existing migrations. If the starter would be deployed, this change would not be made.

Create a new migration and also make it fail-safe by truncating the table before.

dkarnutsch (changes requested):

api/src/db/migrations/Migration20220721122758.ts
jomunker: https://github.com/vivid-planet/comet-starter/pull/315#discussion_r1705171256
Please discuss this with @johnnyomair and let me know what I should do ^^

johnnyomair: I'm sorry, this was my suggestion 😅 We did this multiple times in the past to avoid having a bunch of migrations when starting a new project. We could change our approach though

jomunker: @dkarnutsch what do you prefer?

site/src/middleware.ts
jomunker: https://github.com/vivid-planet/comet-starter/pull/315/commits/a6c7eb999732244667e62d7ff6612d677d86230f

api/src/db/migrations/Migration20220721122758.ts
johnnyomair: @nsams what do you think? Maybe we should start adding actually working migrations once the Starter has been deployed

nsams: A growing Number of Migrations in New Projects Is Not desirable, that is unnecessarily confusing.

Maybe we can drop database and run fixtures on every starter deploy?

johnnyomair: @dkarnutsch what do you think?

dkarnutsch: @johnnyomair 
> Maybe we can drop database and run fixtures on every starter deploy?

I'm okay with that (out of scope for this PR)

dkarnutsch (dismissed):

site/src/middleware.ts
```diff
@@ -70,22 +71,26 @@ export async function middleware(request: NextRequest) {
...
     }
 
+    const scope = { domain: siteConfig.scope.domain };
+
     if (pathname.startsWith("/dam/")) {
         return NextResponse.rewrite(new URL(`${process.env.API_URL_INTERNAL}${request.nextUrl.pathname}`));
     }
 
-    const redirects = await createRedirects();
+    if (scope) {
```
johnnyomair, line 80: Why do you need this if? This will always be true

site/src/middleware.ts
```diff
@@ -101,7 +106,7 @@ export async function middleware(request: NextRequest) {
 
 type RewritesMap = Map<string, Rewrite>;
 
-async function createRewrites(): Promise<RewritesMap> {
+async function createRewrites(scope: Pick<GQLRedirectScope, "domain">): Promise<RewritesMap> {
```
johnnyomair, line 109:
```suggestion
async function createRewrites(scope: GQLRedirectScope): Promise<RewritesMap> {
```

Pick shouldn't be necessary here.

site/src/middleware.ts
jomunker: https://github.com/vivid-planet/comet-starter/pull/315/commits/6c2e52de285641c8dda7543f18254c0b982ea2cc

site/src/middleware.ts
jomunker: https://github.com/vivid-planet/comet-starter/pull/315/commits/7fb53030b9e33dfe2d8b7f4b6ff1be5dfd05cb34

johnnyomair (dismissed):

site/src/redirects/redirects.ts
nsams: I'd rather leave the JSON.stringify, less likely thinks break when scope parts change

nsams (dismissed):

dkarnutsch (approved):

site/src/redirects/redirects.ts
```diff
@@ -113,9 +115,12 @@ const createApiRedirects = async (): Promise<Map<string, Redirect>> => {
...
     return memoryCache.wrap(key, async () => {
-        return new Map<string, Redirect>([...Array.from(await createApiRedirects()), ...Array.from(await createInternalRedirects())]);
+        return new Map<string, Redirect>([
+            ...Array.from(await createApiRedirects({ domain: scope.domain })),
```
johnnyomair, line 122:
```suggestion
            ...Array.from(await createApiRedirects(scope)),
```

thomasdax98 (dismissed):

johnnyomair (approved):

PR #316, opened 2024-07-30
Make scope config a function

thomasdax98:
depends on #283

---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-988


site-configs/main.ts
```diff
@@ -1,10 +1,16 @@
...
-    domains: {
-        local: "localhost:3000",
-    },
-} satisfies Config;
+export default ((env) => {
+    return {
+        name: "Starter Main",
+        domains: {
+            main: "localhost:3000",
+        },
```
thomasdax98, line 8: @nsams 
> 1. the site config does have a single domain (for a given env), I suppose this must not be an object
> 2. you should also define production domain (and other envs)
> 3. It is confusing to have public.domain = "main" vs. domains = "localhost:3000"

thomasdax98: > 1. the site config does have a single domain (for a given env), I suppose this must not be an object

```ts
domains: {
    main: string;
    preliminary?: string;
    pattern?: string;
    additional?: string[];
};

// https://github.com/vivid-planet/comet/blob/de82d79b6055e2eb03f4fb150f83b80732d19e61/packages/cli/src/site-configs.types.ts#L3-L8
```

There could still be a domain pattern or additional domains. I just didn't want to overcomplicate the example in the starter

> 2. you should also define production domain (and other envs)

So I should add an example like this?

```
comet-dxp.com
staging.comet-dxp.com
test.comet-dxp.com
dev.comet-dxp.com
```

> 3. It is confusing to have public.domain = "main" vs. domains = "localhost:3000"

Yes true. I could make it `public.contentScope.domain`. But then I have to put `languages` under `public.contentScope` too and we decided against that previously.

site-configs.d.ts
```diff
@@ -8,8 +8,10 @@ export type ContentScope = {
 export interface SiteConfig extends BaseSiteConfig {
     preloginPassword?: string;
     public: {
-        domain: string;
-        languages: string[];
+        scope: {
+            domain: string;
+            languages: string[];
+        };
```
johnnyomair, line 14: Alternatively we could make a `scopes: Array<{ domain: string, language: string }>` array

johnnyomair (dismissed):

site-configs/secondary.ts
```diff
@@ -1,10 +1,26 @@
-import { Config } from "../site-configs";
+import {Environment, GetSiteConfig} from "../site-configs";
```
nsams, line 1:
```suggestion
import { Environment, GetSiteConfig } from "../site-configs";
```
seems like prettier (and everyting related eslint/pipeline etc) isn't working

site-configs.d.ts
nsams: but that doesn't allow using `public.scope.domain`

nsams (approved):

nsams (approved):

nsams (approved):

johnnyomair (approved):

PR #323, opened 2024-08-06
Simplify postgres-password handling

fraxachun:
- Remove encoding as it is not necessary
- Avoid abbreviation


kaufmo (approved):

dkarnutsch (approved):
As evaluated by @Schreglmann this makes problems if the password includes `“’()`, but I think this is something we can ignore as e.g. 1Password never uses these characters for creating passwords

thomasdax98 (approved):

johnnyomair (approved):

nsams (approved):

PR #324, opened 2024-08-06
Digitalocean

dkarnutsch:
- [ ] Buildpacks should install dev dependencies as well
- [ ] site/comet-config.json does not work if symlink (at runtime, buildtime is fine)

---

## PR Checklist

-   [ ] Link to the respective task if one exists: 
-   [ ] Provide screenshots/screencasts if the change contains visual changes

<details>
    
</details>


api/src/db/ormconfig.ts
```diff
@@ -10,7 +10,7 @@ export const ormConfig = createOrmConfig({
     password: process.env.POSTGRESQL_PASSWORD,
     dbName: process.env.POSTGRESQL_DB,
     driverOptions: {
-        connection: { ssl: process.env.POSTGRESQL_USE_SSL === "true" },
+        connection: { ssl: process.env.POSTGRESQL_USE_SSL === "true" ? { ca: process.env.POSTGRESQL_CA_CERT } : undefined },
```
johnnyomair, line 13: Out of scope: Why don't we use the config here instead of process.env?

admin/block-meta.json
johnnyomair, line 1: The symlinks shouldn't be committed in the final version 😁

admin/package.json
```diff
@@ -6,6 +6,7 @@
         "serve": "cd server && npm run serve",
         "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -e .env -e .env.local -e .env.secrets -e .env.site-configs -- vite",
         "build": "npm run intl:compile && run-p gql:types generate-block-types && NODE_ENV=production vite build",
+        "build:digitalocean": "./intl-update.sh && npm install --production=false && npm run build && npm ci",
```
johnnyomair, line 9: Why this duplicate npm install?

api/src/db/ormconfig.ts
dkarnutsch: addressed in #365

dkarnutsch:
Continued in #368

PR #329, opened 2024-08-08
Add `IntlProvider` to block preview

SebiVPS:
We sometimes need translated texts in blocks. We need to provide an IntlProvider for this to work in the block preview. In this PR, the block preview page is moved to be located under `/[language]/block-preview` to use the IntlProvider defined in `/[language]/layout.tsx`.


johnnyomair (approved):

nsams:
Now https://github.com/vivid-planet/comet-starter/blob/main/site/src/middleware.ts#L39 doesn't match anymore.

johnnyomair:
> Now https://github.com/vivid-planet/comet-starter/blob/main/site/src/middleware.ts#L39 doesn't match anymore.

Could we remove that? Why do we even need it?

nsams:
> Could we remove that? Why do we even need it?

I think that is needed, because for block preview the draft mode is not entered (like it is for site preview) and so the getSiteConfigForHost() doesn't match a domain (as we are at the preview domain) and also doesn't find draft mode (as we didn't enter it) resulting in "Cannot get siteConfig for host ${host}".

This doesn't hapen locally because the preview domain == main domain.

One solution could be to look for preview domain instead of `/block-preview`?

nsams:
> What do you prefer?

making any site preview path that includes `/block-preview/` in it's path public is not an option, so we need the regex. Or, alternatively we find a different solution to inject the language (scope thru iframe bridge?) and use always a fixed path `domain/block-preview/`

johnnyomair:
@fraxachun I don't think we can close this since #349 has never been merged and #351 doesn't add the change made in this PR.

johnnyomair:
@fraxachun I just had a look at this again and I don't think it's easily possible to achieve this with the iframebridge's content scope. I'd need to load all messages of all languages and then get the correct locale on the client (like you do for site configs). Do you think this is better than moving the block-preview to the language path?

johnnyomair:
Just tried this again on the latest `main`. It seems that this was fixed by #357.

PR #330, opened 2024-08-08
Remove uninstall.sh script

johnnyomair:
The script is frequently used to try to fix probably unrelated errors, e.g., a failing migration. We shouldn't promote this way of "debugging", so I'd suggest to remove the script altogether. If developers really need to "uninstall" _everything_, they'll find the commands themselves.


dkarnutsch (approved):
An alternative would be to add a clear warning that is not the recommend way, similar to NPM:

>➜  ~/ws/comet-v7 git:(docs/enhance-cron-job-module) npm cache clear
npm error As of npm@5, the npm cache self-heals from corruption issues
npm error   by treating integrity mismatches as cache misses.  As a result,
npm error   data extracted from the cache is guaranteed to be valid.  If you
npm error   want to make sure everything is consistent, use `npm cache verify`
npm error   instead.  Deleting the cache can only make npm go slower, and is
npm error   not likely to correct any problems you may be encountering!
npm error
npm error   On the other hand, if you're debugging an issue with the installer,
npm error   or race conditions that depend on the timing of writing to an empty
npm error   cache, you can use `npm install --cache /tmp/empty-cache` to use a
npm error   temporary cache instead of nuking the actual one.
npm error
npm error   If you're sure you want to delete the entire cache, rerun this command
npm error   with --force.

nsams (approved):

johnnyomair:
> An alternative would be to add a clear warning that is not the recommend way, similar to NPM

I believe this will just result in devs changing their muscle memory to `./uninstall.sh --force`. Or would you add a prompt in the script itself?

dkarnutsch:
> I believe this will just result in devs changing their muscle memory to `./uninstall.sh --force`

You are probably right... just remove it 👍

thomasdax98 (approved):

PR #335, opened 2024-08-13
Add `StandaloneCallToActionListBlock`

SebiVPS:
Add `StandaloneCallToActionListBlock` which contains alignment.
We discussed that we need to align the call to action list block items.

### DIFF
--- a/admin/src/common/blocks/StandaloneCallToActionListBlock.tsx
+++ b/admin/src/common/blocks/StandaloneCallToActionListBlock.tsx
@@ -1,0 +1,35 @@
+import { BlockCategory, createCompositeBlock, createCompositeBlockSelectField } from "@comet/blocks-admin";
+import { StandaloneCallToActionListBlockData } from "@src/blocks.generated";
+import { CallToActionListBlock } from "@src/common/blocks/CallToActionListBlock";
+import { FormattedMessage } from "react-intl";
+
+export const StandaloneCallToActionListBlock = createCompositeBlock(
+    {
+        name: "StandaloneCallToActionList",
+        displayName: <FormattedMessage id="standaloneCallToActionList.displayName" defaultMessage="CallToActionList" />,
+        blocks: {
+            callToActionList: {
+                block: CallToActionListBlock,
+            },
+            alignment: {
+                block: createCompositeBlockSelectField<StandaloneCallToActionListBlockData["alignment"]>({
+                    defaultValue: "left",
+                    options: [
+                        { value: "left", label: <FormattedMessage id="standaloneCallToActionList.alignment.left" defaultMessage="left" /> },
+                        { value: "center", label: <FormattedMessage id="standaloneCallToActionList.alignment.center" defaultMessage="center" /> },
+                        { value: "right", label: <FormattedMessage id="standaloneCallToActionList.alignment.right" defaultMessage="right" /> },
+                    ],
+                    fieldProps: {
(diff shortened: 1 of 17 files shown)


site/src/common/blocks/CallToActionListBlock.tsx
```diff
@@ -17,14 +16,6 @@ export const CallToActionListBlock = withPreview(
...
 );
 
-export const PageContentCallToActionListBlock = (props: CallToActionListBlockProps) => (
-    <PageLayout grid>
-        <PageLayoutContent>
-            <CallToActionListBlock {...props} />
-        </PageLayoutContent>
-    </PageLayout>
-);
-
```
johnnyomair, line 27: Great that you've thought about removing this now unneeded code 👏🏼

site/src/common/blocks/StandaloneCallToActionListBlock.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <PageLayout grid>
+        <PageLayoutContent>
+            <StandaloneCallToActionListBlock {...props} />
+        </PageLayoutContent>
+    </PageLayout>
+);
+
+const Root = styled.div<{ $alignment: CSSProperties["justifyContent"] }>`
+    display: flex;
+    justify-content: ${({ $alignment }) => $alignment};
```
johnnyomair, line 32: Shouldn't this be flex-start and flex-end instead of start and end?

SebiVPS: interesting, on MDN Web Docs it says start, end and left, right are possible values.
on w3schools with flex-start and flex-end, 
it all does work.

I added a map for flex-start and flex-end values.

johnnyomair: @jamesricky what's your opinion? IMO we can implement it without mapping if start and end are valid. 

All my CSS knowledge is based around 2017 😁

jamesricky: I'd prefer `flex-start`/`flex-end` as it seems to have slightly better browser support. 
But I'm fine with both 😄

admin/src/common/blocks/StandaloneCallToActionListBlock.tsx
```diff
@@ -0,0 +1,35 @@
...
+            callToActionList: {
+                block: CallToActionListBlock,
+            },
+            alignment: {
+                block: createCompositeBlockSelectField<StandaloneCallToActionListBlockData["alignment"]>({
+                    defaultValue: "left",
+                    options: [
+                        { value: "left", label: <FormattedMessage id="standaloneCallToActionList.alignment.left" defaultMessage="left" /> },
+                        { value: "center", label: <FormattedMessage id="standaloneCallToActionList.alignment.center" defaultMessage="center" /> },
+                        { value: "right", label: <FormattedMessage id="standaloneCallToActionList.alignment.right" defaultMessage="right" /> },
```
jamesricky, line 20: Probably out of scope for this PR, but should we add arrays for `leftRightOptions` and `leftRightCenterOptions` to some kind of common/utils file, as this is quite often needed? 🤔 

Then we also wouldn't need unique message IDs which would make translation slightly easier.

johnnyomair: The messages should probably be defined in [messages.ts](https://github.com/vivid-planet/comet/blob/main/packages/admin/admin/src/messages.ts)

johnnyomair (approved):

PR #337, opened 2024-08-13
Remove barrel React imports

SebiVPS:
Use named imports instead of barrel importing React. Doing so will result in less code and better readability. The new [React docs](https://react.dev/) also use named imports and never import React itself.


site/src/common/helpers/HiddenIfInvalidLink.tsx
```diff
@@ -7,11 +7,10 @@ import {
     LinkBlockData,
     PhoneLinkBlockData,
 } from "@src/blocks.generated";
-import * as React from "react";
+import { PropsWithChildren } from "react";
 
-interface HiddenIfInvalidLinkProps {
+interface HiddenIfInvalidLinkProps extends PropsWithChildren {
```
johnnyomair, line 12: This should probably be `type HiddenIfInvalidLinkProps = PropsWithChildren<{ link: LinkBlockData }>` to correctly use the generic. You could also add this in [L16](https://github.com/vivid-planet/comet-starter/pull/337/files#diff-37796abfa0bf59d75e601d0b7d2ff9eba664ac4491f5ee76a8c8c8d31589a8ccR16)

johnnyomair (commented):
We should have an ESLint rule for this

dkarnutsch:
I would prefer to use the same rules for API and Admin

johnnyomair:
> I would prefer to use the same rules for API and Admin

API doesn't have React, what do you mean? 🤔

dkarnutsch:
> > I would prefer to use the same rules for API and Admin
> 
> API doesn't have React, what do you mean? 🤔

Admin and Site, sorry...

johnnyomair:
> > > I would prefer to use the same rules for API and Admin
> > 
> > 
> > API doesn't have React, what do you mean? 🤔
> 
> Admin and Site, sorry...

Yes, we should get rid of those imports everywhere.

johnnyomair:
@SebiVPS there are still some `import React from "react"` imports, and some `React.` usages without import (I don't know how this even works). Could you please remove them as well?

dkarnutsch (approved):

admin/.eslintrc.json
```diff
@@ -2,6 +2,17 @@
...
+            "error",
+            {
+                "paths": [
+                    {
+                        "name": "react",
+                        "importNames": ["default"]
+                    }
+                ]
+            }
+        ]
```
johnnyomair, line 16: We should move both this rules to `@comet/eslint-config` in v8

johnnyomair (approved):

admin/.eslintrc.json
thomasdax98, line 16: I think this might still crash the admin generator 🤔

SebiVPS: yep, when its used.
What do you suggest? 
You think we can remove the barrel imports in the generator files without breaking other projects? Or do we need to wait for v8?

SebiVPS: @thomasdax98 @johnnyomair 
The restrict import rule is now just added for the site/ until v8 👍 due to the admin/generator

johnnyomair (approved):

PR #344, opened 2024-08-26
Fix Rerendering Bug in MasterMenu

thomasdax98:
## Previously

The MasterMenu was reset completely on every site change, closing open collapsibles

https://github.com/user-attachments/assets/83ee4a75-8d89-4373-a74f-e213c60a15c0

## Now

The MasterMenu is not reset anymore. Open collapsibles stay open

https://github.com/user-attachments/assets/05af2c39-7196-454d-a363-709d8d844773

---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-1042

### DIFF
--- a/admin/src/common/MasterMenu.tsx
+++ b/admin/src/common/MasterMenu.tsx
@@ -4,6 +4,7 @@ import {
     ContentScopeIndicator,
     createRedirectsPage,
     DamPage,
+    MasterMenu as CometMasterMenu,
     MasterMenuData,
     PagesPage,
     PublisherPage,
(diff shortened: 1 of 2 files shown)


admin/src/common/MasterMenu.tsx
```diff
@@ -102,3 +103,5 @@ export const masterMenuData: MasterMenuData = [
         requiredPermission: "pageTree",
     },
 ];
+
+export const MasterMenu = () => <CometMasterMenu menu={masterMenuData} />;
```
nsams, line 107: Having a component named the same in application and library is not ideal.

you could name it also the other way around:
`export const AppMasterMenu = () => <MasterMenu menu={masterMenuData} />;`

thomasdax98: 73072eef4af447eb3b13253a55d432bb861636ad

nsams (approved):

PR #347, opened 2024-08-28
Add mobile header

SebiVPS:
Add header styling according to Comet Starter Figma design. Also take accessibility into consideration.

---

## PR Checklist

-   [x] Link to the respective task if one exists: COM-1046

<details>

https://github.com/user-attachments/assets/30f5cfca-526b-48c3-a626-fb81147e5f2d

</details>


site/src/layout/header/MobileHeaderMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+            setExpandedSubLevelNavigation(null);
+        } else {
+            setExpandedSubLevelNavigation(id);
+        }
+    };
+
+    // prevent scrolling when menu is open
+    useEffect(() => {
+        document.body.style.overflow = isMenuOpen ? "hidden" : "unset";
+    }, [isMenuOpen]);
```
johnnyomair, line 34: In our projects we have a `const DisableBodyScrolling = createGlobalStyle...` helper for this, should we do the same here?

site/src/layout/header/MobileHeaderMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+                                            <sc.Link page={node} aria-label={node.name}>
+                                                {node.name}
+                                            </sc.Link>
+                                        )}
+                                        {node.childNodes.length > 0 && (
+                                            <sc.SubLevelNavigation $isExpanded={expandedSubLevelNavigation === node.id}>
+                                                <PageLayout grid>
+                                                    <sc.PageLayoutContent>
+                                                        <li>
+                                                            <sc.BackButton aria-label="Go back" onClick={() => setExpandedSubLevelNavigation(null)}>
```
johnnyomair, line 69: This should be translatable

site/src/layout/header/MobileHeaderMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+            <sc.MenuContainer $isMenuOpen={isMenuOpen} aria-hidden={!isMenuOpen}>
+                <PageLayout grid>
+                    <sc.PageLayoutContent>
+                        <nav>
+                            <sc.TopLevelNavigation>
+                                {header.map((node) => (
+                                    <li key={node.id}>
+                                        {node.childNodes.length > 0 ? (
+                                            <sc.ButtonLink
+                                                aria-label={`Submenu of ${node.name}`}
```
johnnyomair, line 50: This should be translatable

site/src/layout/header/Header.tsx
```diff
@@ -196,4 +301,151 @@ const Link = styled(PageLink)`
...
+    color: inherit;
+    padding: 0;
+    width: 24px;
+    height: 24px;
+`;
+
+const IconWrapper = styled.div`
+    width: 16px;
+    height: 16px;
+    color: inherit;
```
jamesricky, line 327: I think that's the default behavior of divs

```suggestion
```

site/src/layout/header/Header.tsx
```diff
@@ -90,6 +100,101 @@ const Header = ({ header }: Props) => {
                                 ))}
                             </TopLevelNavigation>
                         </DesktopHeaderFullHeightNav>
+
+                        <MobileMenuContainer>
```
jamesricky, line 104: Would it make sense to move the mobile menu into it's own component?
Might improve readability and you wouldn't have to prefix all the elements with "Mobile"?

site/src/layout/header/Header.tsx
```diff
@@ -196,4 +301,151 @@ const Link = styled(PageLink)`
...
+const MenuContainer = styled.div<{ $isMenuOpen: boolean }>`
+    display: block;
+    position: fixed;
+    top: ${headerHeight}px;
+    left: 0;
+    height: 0;
+    width: 100vw;
+    z-index: 40;
+    background-color: ${({ theme }) => theme.palette.gray["200"]};
+    overflow: hidden;
```
jamesricky, line 345: What happens if there are more items than there is space for?
You should probably be able to scroll the items.

site/src/layout/header/Header.tsx
```diff
@@ -196,4 +301,151 @@ const Link = styled(PageLink)`
...
+    justify-content: space-between;
+    width: 100%;
+    padding: ${({ theme }) => theme.spacing.S500} 0;
+
+    &:hover {
+        color: ${({ theme }) => theme.palette.primary.main};
+    }
+`;
+
+const BackButton = styled.button`
```
jamesricky, line 417: Would it make sense to extend `ButtonLink`, as the styles are almost the same?

site/src/util/DisableBodyScrolling.ts
```diff
@@ -0,0 +1,9 @@
+import { createGlobalStyle } from "styled-components";
+
+export const DisableBodyScrolling = createGlobalStyle`
+    html {
+        overflow: hidden;
+        overscroll-behavior: contain;
```
jamesricky, line 6: Why exactly is this needed? 🤔

site/src/layout/header/Header.tsx
```diff
@@ -1,21 +1,31 @@
...
 }
 
+const headerHeight = 100;
+
 const Header = ({ header }: Props) => {
     const intl = useIntl();
+    const [isMenuOpen, setIsMenuOpen] = useState<boolean>(false);
     const [expandedSubLevelNavigation, setExpandedSubLevelNavigation] = useState<string | null>(null);
 
+    const toggleMenu = () => {
```
johnnyomair, line 24: Handlers usually start with handle, like `handleMenuToggle` or `handleMenuButtonClick`

site/src/layout/header/Header.tsx
```diff
@@ -1,21 +1,31 @@
...
+        if (isMenuOpen) setExpandedSubLevelNavigation(null);
+        setIsMenuOpen((prev) => !prev);
```
johnnyomair, line 26:
```suggestion
        if (isMenuOpen) {
            setExpandedSubLevelNavigation(null);
            setIsMenuOpen(false);
        } else {
            setIsMenuOpen(true);    
        }
```

IMO this is more readable

site/src/layout/header/Header.tsx
```diff
@@ -90,6 +100,101 @@ const Header = ({ header }: Props) => {
...
+                                onClick={toggleMenu}
+                            >
+                                <Icon href={isMenuOpen ? "/assets/icons/menu-open.svg#menu-open" : "/assets/icons/menu.svg#menu"} />
+                            </MenuButton>
+                            <MenuContainer $isMenuOpen={isMenuOpen} aria-hidden={!isMenuOpen}>
+                                <PageLayout grid>
+                                    <PageLayoutContent>
+                                        <nav>
+                                            <MobileTopLevelNavigation>
+                                                {header.map((node) => (
```
johnnyomair, line 121: We're currently not considering [hideInMenu](https://github.com/vivid-planet/comet-starter/blob/main/api/schema.gql#L156), right? This should be changed.

site/src/layout/header/Header.tsx
johnnyomair: I recommended otherwise, but I wasn't aware that the complete rendering logic of the page tree nodes needs to be duplicated. So maybe it makes sense after all.

However, the MobileMenu component mustn't reuse the GraphQL fragment from Header, but rather define its own.

site/src/layout/header/Header.tsx
SebiVPS: There would be a earlier (with less items) a Problem in the desktop header menu.  How should we handle that?

site/src/util/DisableBodyScrolling.ts
SebiVPS: @johnnyomair said we should copy that from existing projects, do we need that?

johnnyomair: I guess @jamesricky is talking about this line in particular. I don't know why that one's needed. 🤷🏼‍♂️ We do need the DisableBodyScrolling helper though 😁

site/src/layout/header/Header.tsx
SebiVPS: ok, however, why can't i reuse the GraphQL Fragment? Should that just be duplicated with another name? I do not know how that should be possible. There is only the HeaderFragment used in the gql query?

johnnyomair: There should be a MobileMenuFragment that queries all fields it requires (probably the same as HeaderFragment). A component should define which data it needs, not rely on a parent loading the correct fields. This technique is called colocating fragments. See https://www.apollographql.com/docs/react/data/fragments/#colocating-fragments

site/src/layout/header/Header.tsx
SebiVPS: Good point 👍

site/src/util/DisableBodyScrolling.ts
SebiVPS: if nobody knows, i delete it 🤷

site/src/layout/header/Header.tsx
```diff
@@ -46,50 +47,54 @@ const Header = ({ header }: Props) => {
...
-                                        key={node.id}
-                                        onMouseEnter={() => setExpandedSubLevelNavigation(node.id)}
-                                        onMouseLeave={() => setExpandedSubLevelNavigation(null)}
-                                    >
-                                        <LinkContainer>
-                                            <Link page={node} activeClassName="active" aria-label={node.name}>
-                                                {node.name}
-                                            </Link>
+                                {header
+                                    .filter((node) => !node.hideInMenu)
```
jamesricky, line 51: Shouldn't this already be excluded from the api response?

site/src/layout/header/Header.tsx
SebiVPS: ok, i set the overflow prop to auto, to make the list scrollable for many items 👍

site/src/layout/header/Header.tsx
jamesricky: Exactly, I think this data should not be returned by the api at all

SebiVPS: actually there is set `excludeHiddenInMenu: true` here
https://github.com/vivid-planet/comet-starter/blob/ad75f6e6b74e3ccc26ee4686b5f45402169848bc/api/src/menus/menus.resolver.ts#L16C91-L16C110

So i removed the client side filter and start investigate, why this does not work anymore.
I guess this will be an library change ... so no blocker for this PR i guess, because this is a bug fix in the library.

site/src/layout/header/Header.tsx
SebiVPS: https://github.com/vivid-planet/comet-starter/pull/347#discussion_r1768403849

site/src/layout/header/Header.tsx
SebiVPS: We found out, that the `excludeHiddenInMenu: true` for `pageTreeRootNodeList` only filters for the first level and is not forwarded to the childNodes (which one would expects here).
@johnnyomair @thomasdax98  can we discuss this next week in person?

johnnyomair: childNodes can't access the argument since it's a field resolver. We'd need the argument for the field resolver as well...

site/src/layout/header/Header.fragment.ts
```diff
@@ -1,4 +1,5 @@
 import { gql } from "@comet/cms-site";
+import { mobileMenuFragment } from "@src/layout/header/MobileMenu.fragment";
```
johnnyomair, line 2: Please use a relative import

site/src/layout/header/MobileMenu.tsx
```diff
@@ -0,0 +1,287 @@
...
+import { SvgUse } from "@src/common/helpers/SvgUse";
+import { GQLMobileMenuFragment } from "@src/layout/header/MobileMenu.fragment.generated";
+import { PageLink } from "@src/layout/header/PageLink";
+import { PageLayout } from "@src/layout/PageLayout";
+import { DisableBodyScrolling } from "@src/util/DisableBodyScrolling";
+import { useEffect, useState } from "react";
+import { FormattedMessage, useIntl } from "react-intl";
+import styled, { css } from "styled-components";
+
+export const HEADER_HEIGHT = 100;
```
johnnyomair, line 12: Why does the mobile menu define the header height? And shouldn't this be a CSS variable?

site/src/layout/header/MobileMenu.tsx
```diff
@@ -0,0 +1,287 @@
...
+            setExpandedSubLevelNavigation(null);
+        } else {
+            setExpandedSubLevelNavigation(id);
+        }
+    };
+
+    useEffect(() => {
+        if (!expandedSubLevelNavigation) return;
+
+        const keyDownHandler = (e: KeyboardEvent) => {
```
johnnyomair, line 43: Naming: handleKeyDown. And please use event instead of e.

site/src/layout/header/MobileMenu.tsx
```diff
@@ -0,0 +1,284 @@
...
+    background-color: ${({ theme }) => theme.palette.gray["200"]};
+    overflow: auto;
+    visibility: hidden;
+    transition: height 0.15s ease-out, visibility 0s linear 0.15s;
+
+    ${({ $isMenuOpen }) =>
+        $isMenuOpen &&
+        css`
+            visibility: visible;
+            height: calc(100vh - var(--header-height));
```
jamesricky, line 197: Have you tested this in iOS Safari? 
Since scrolling the page shows/hides the address-bar, this can sometimes lead to issues when using `vh`, see: https://www.sabhya.dev/handling-ios-safari-toolbar-for-full-height-web-content

site/src/layout/header/MobileMenu.tsx
```diff
@@ -0,0 +1,284 @@
...
+`;
+
+const SubLevelNavigation = styled.ol<{ $isExpanded: boolean }>`
+    list-style-type: none;
+    position: fixed;
+    top: var(--header-height);
+    left: 100vw;
+    height: calc(100vh - var(--header-height));
+    width: 100vw;
+    z-index: 41;
```
jamesricky, line 215: Since this is already nested inside `MenuContainer`, wouldn't it be enough to just span the full width and height? 🤔 

```suggestion
    position: absolute;
    inset: 0;
    z-index: 1; // Maybe not even necessary
```

site/src/layout/header/MobileMenu.tsx
```diff
@@ -0,0 +1,284 @@
...
+    padding: ${({ theme }) => theme.spacing.S500} 0;
+    text-decoration: none;
+    color: ${({ theme }) => theme.palette.text.primary};
+
+    &:hover {
+        color: ${({ theme }) => theme.palette.primary.main};
+    }
+`;
+
+export { MobileMenu };
```
jamesricky, line 284: I don't think having an extra line for a single export is beneficial... 
I'd just always export the variable/function directly at the declaration.

site/src/layout/header/MobileMenu.tsx
SebiVPS: 🤔  you missed the transition / animation from the right side, that does not work with position: absolute;

johnnyomair (approved):
LGTM.

site/src/layout/header/MobileMenu.tsx
jamesricky: With either an iPhone or an iPhone simulator 🙂

site/src/layout/header/MobileMenu.tsx
jamesricky: You can use `transformX` for the animation, that's probably smoother than `left` anyway

site/src/layout/header/MobileMenu.tsx
SebiVPS: Unfortunately, I was not successful with either of them
Screenshot 2024-09-27 at 07 49 20
Simulator Screenshot - iPhone 16 Pro - 2024-09-27 at 07 24 35

site/src/layout/header/MobileMenu.tsx
SebiVPS: in the chrome device dev tools the screen is extended (scrollable to right) when position absolute is used

site/src/layout/header/MobileMenu.tsx
jamesricky: Accessing `http://localhost:3000` should always work in the simulator, if `npm run dev` is running on the same device as the simulator 🤔 

Or is the issue that some things don't work in Safari in general, due to the security policy bug (COM-864)?

site/src/layout/header/MobileMenu.tsx
jamesricky: You probably need to set `overflow: hidden` on some parent then... but if it works fine with `fixed` then I'm fine with that too 🙂

SebiVPS: overflow is set to auto on the parent due to https://github.com/vivid-planet/comet-starter/pull/347#discussion_r1760641149

site/src/layout/header/Header.tsx
SebiVPS: i re-added the client side filter for now

site/src/layout/header/MobileMenu.tsx
SebiVPS: Still no luck on this.

jamesricky: I tried it in the iOS 18 simulator locally unfortunately there are two issues, see the video below:

- The last item cannot be scrolled to, most likely due to the issue I mentioned [above](https://github.com/vivid-planet/comet-starter/pull/347#discussion_r1776639984)
- When scrolling in the sub menu sometimes the wrong element will be scrolled (starting at second 10)

https://github.com/user-attachments/assets/0144894f-0b87-4141-aeab-9fbb55f1381e

SebiVPS: Ok, thanks for the info. 
I could fix the issue with the height.
How would you handle the highlighting? That must come from the :hover prop thats gonna triggered, when you lift up the finger on that element.

jamesricky (approved):

johnnyomair (approved):

site/src/layout/header/MobileMenu.tsx
```diff
@@ -0,0 +1,286 @@
...
+                        </nav>
+                    </PageLayoutContent>
+                </PageLayout>
+            </MenuContainer>
+        </Root>
+    );
+};
+
+const Root = styled.div`
+    ${({ theme }) => css`
```
mariokemetinger, line 151: Why is this written like that? Could be shorter:
`${({ theme }) => theme.breakpoints.sm.mediaQuery} {
    display: none;
}`

mariokemetinger (approved):

PR #348, opened 2024-08-30
Prepare AuthModule for AuthProxy

fraxachun:
Since the variables are already defined we can also use them. This puts the AuthModule closer to a production-ready setup.

---

## PR Checklist

-   [ ] Link to the respective task if one exists: 
-   [ ] Provide screenshots/screencasts if the change contains visual changes

<details>
    
</details>


api/src/app.module.ts
```diff
@@ -86,10 +87,10 @@ export class AppModule {
                         ),
                         userService,
                         accessControlService,
-                        systemUsers: ["system"],
+                        systemUsers: ["system-user"],
```
thomasdax98, line 90: I believe the user name in the AccessLogModule (line 128) should also be "system-user" now. Maybe we should even put this into a constant since it's now hardcoded in multiple places in the API

site/src/util/graphQLClient.ts
```diff
@@ -13,7 +13,7 @@ export function createGraphQLFetch(previewData?: SitePreviewData) {
     };
 
     if (isServerSide) {
-        headers.authorization = `Basic ${Buffer.from(`vivid:${process.env.API_PASSWORD}`).toString("base64")}`;
+        headers.authorization = `Basic ${Buffer.from(`system-user:${process.env.API_BASIC_AUTH_PASSWORD}`).toString("base64")}`;
```
thomasdax98, line 16: Is that correct? Here you use `API_BASIC_AUTH_PASSWORD` but in the env variables it's `BASIC_AUTH_PASSWORD`:
https://github.com/vivid-planet/comet-starter/blob/b0426eb6cea05e27f1edb4cef4e93d200316911f/api/src/config/environment-variables.ts#L44

I believe they should be identical

thomasdax98 (dismissed):

site/src/util/graphQLClient.ts
nsams: A better name might be `API_SYSTEM_USER_PASSWORD`, the important thing is not how it's used (basic auth), rather than what it is used for (to authorize the system-user)

johnnyomair: I tend to agree with @nsams, but I don't have a strong opinion about this.

How about API_BASIC_AUTH_SYSTEM_USER_PASSWORD? 😉

api/src/auth/auth.module.ts
```diff
@@ -1,26 +1,67 @@
...
+                }),
+            );
+            providers.push({
+                provide: APP_GUARD,
+                useClass: createCometAuthGuard(["auth-proxy-jwt", "system-user"]),
+            });
+        } else {
+            if (process.env.NODE_ENV !== "development") {
+                throw new Error("config.auth.useAuthproxy must only be false in development");
+            }
```
johnnyomair, line 49: We could also validate this in environment-variables.ts

site/src/util/SiteConfig.ts
johnnyomair, line 1: IMO the file should be cased siteConfig, since it's not a React component

api/src/auth/auth.module.ts
nsams: if it is too complicated only here is also Ok

johnnyomair: I believe this would work:

```ts
@Transform(({ value }) => value === "true")
@ValidateIf(() => process.env.NODE_ENV === "production")
@Equals(true)
USE_AUTHPROXY: boolean = false;
```

But I'm not sure if I like it 😅 So maybe we should leave it like it is.

nsams (approved):

johnnyomair (approved):

thomasdax98 (approved):

PR #351, opened 2024-09-03
Improve rendering of site

fraxachun:
- Remove `next-runtime-env` as it prevents static rendering (use context in `layout.tsx` instead)
- Get rid of `config.ts`
- Simplify block preview and provide site config context in there as well
- Use separate `layout.tsx` files for block preview and site

Supersedes https://github.com/vivid-planet/comet-starter/pull/350 and https://github.com/vivid-planet/comet-starter/pull/349

To discuss:
- Move `SiteConfigsProvider.tsx` and `SiteConfigProvider.tsx` to @comet/cms-site?
- Shall we document the following explanation and if yes - where?

----

**Ways to access the current SiteConfig**
- _Page_
Use `await getSiteConfigForDomain(params.domain)`
- _Component_ (Server Rendering)
Not possible, instead pass the necessary data from the page
- _Component or Block-Preview_ (Client Rendering)
Use the `useSiteConfig()`-hook


site/package.json
```diff
@@ -28,7 +28,6 @@
         "cache-manager": "^5.6.1",
         "graphql": "^15.0.0",
         "next": "^14.2.0",
-        "next-runtime-env": "^3.2.2",
```
thomasdax98, line 31: For information purposes:
Another problem of next-runtime-env is that it doesn't work correctly in the not-found page, see:

- https://github.com/expatfile/next-runtime-env/issues/147
- https://github.com/expatfile/next-runtime-env/issues/106
- https://github.com/vercel/next.js/issues/63980

site/src/app/[domain]/layout.tsx
```diff
@@ -1,33 +1,34 @@
...
 
 export default async function RootLayout({
     children,
+    params: { domain },
 }: Readonly<{
     children: React.ReactNode;
+    params: { domain: string };
 }>) {
-    const { gtmId } = await getSiteConfig();
+    const siteConfig = await getSiteConfigForDomain(decodeURIComponent(domain));
```
Comment (thomasdax98) [inline site/src/app/[domain]/layout.tsx:21]: Do we even need the `decodeURIComponent`?

site/src/app/block-preview/[type]/page.tsx
```diff
@@ -31,14 +35,37 @@ const PreviewPage: React.FunctionComponent = () => {
...
+    return stage ? <StageBlock data={stage} /> : null;
+};
+
+const previewComponents = {
+    page: PagePreview,
+    stage: StagePreview,
+};
+
+const PreviewWrapper = ({ type }: { type: string }) => {
+    const siteConfig = useBlockPreviewSiteConfig((siteConfigs, contentScope) =>
```
Comment (thomasdax98) [inline site/src/app/block-preview/[type]/page.tsx:50]: Shouldn't we define the `getSiteConfigFromContentScope` method once in `useBlockPreviewSiteConfig`?

We only use the hook once atm, but it seems unnecessary to me that you have to define this every time you use the hook

thomasdax98 (dismissed):

site/src/middleware.ts
```diff
@@ -22,6 +28,16 @@ async function getSiteConfigForHost(host: string) {
     return getSiteConfigs().find((siteConfig) => siteConfig.domains.main === host || siteConfig.domains.preliminary === host);
 }
 
+let siteConfigs: PublicSiteConfig[];
+export function getSiteConfigs() {
+    if (!siteConfigs) {
+        const json = process.env.NEXT_PUBLIC_SITE_CONFIGS;
+        if (!json) throw new Error("process.env.NEXT_PUBLIC_SITE_CONFIGS must be set.");
```
nsams, line 35: this with the removal of next-runtime-env this should not be a NEXT_PUBLIC_ anymore, as it can't be used on client side anymore

site/src/app/[domain]/[language]/[[...path]]/page.tsx
```diff
@@ -1,3 +1,5 @@
+export const dynamic = "error";
```
Comment (thomasdax98) [inline site/src/app/[domain]/[language]/[[...path]]/page.tsx:1]: I just noticed that this makes the SitePreview break:

Bildschirmfoto 2024-09-05 um 09 27 43

nsams (approved):

thomasdax98 (approved):

site/src/app/block-preview/[type]/page.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const StagePreview = () => {
+    const { block: stage } = useIFrameBridge();
+    return stage ? <StageBlock data={stage} /> : null;
+};
+
+const previewComponents = {
+    page: PagePreview,
+    stage: StagePreview,
+};
```
Comment (johnnyomair) [inline site/src/app/block-preview/[type]/page.tsx:46]: Should we move this preview components into separate files? Otherwise this file may get rather large when having many different previews

PR #353, opened 2024-09-10
Split block-preview pages

fraxachun:
Simplifies adding new block-preview components


johnnyomair (approved):
I was thinking about something like [documentTypes](https://github.com/vivid-planet/comet-starter/blob/main/site/src/documents/index.tsx#L13) (only the components in separate files), but this is also nice

nsams (approved):

PR #354, opened 2024-09-10
Block preview intl

fraxachun:
Depends on https://github.com/vivid-planet/comet-starter/pull/353

Supersedes https://github.com/vivid-planet/comet-starter/pull/329


site/src/app/block-preview/[language]/layout.tsx
```diff
@@ -0,0 +1,12 @@
+import { loadMessages } from "@src/app/[domain]/[language]/layout";
+import { IntlProvider } from "@src/util/IntlProvider";
+import { PropsWithChildren } from "react";
+
+export default async function Page({ children, params: { language } }: PropsWithChildren<{ params: { language: string } }>) {
```
Comment (johnnyomair) [inline site/src/app/block-preview/[language]/layout.tsx:5]: This needs to be duplicated because of the language param being after /block-preview. Don't you think it's better to move the param before it and change the if in the middleware?

nsams:
using the language from the iframebridge injected scope didn't work?

nsams:
> only work if we loaded all translations

could be done (if it's for block preview only)

> Thinking about that I'm tending towards a solution where the site-configs are in the path as well. This makes the block-preview more similar to the normal pages.

that also sounds good. We might even remove the scope passing from the iframe bridge then?

PR #355, opened 2024-09-12
Add script to build and run the site to simulate deployment

fraxachun:
https://github.com/user-attachments/assets/f0db8c39-0803-410c-b788-422b5663eb6a


build-and-run-site.sh
```diff
@@ -0,0 +1,19 @@
...
+
+echo "[1/3] Stop site if running..."
+npx dev-pm stop site
+echo ""
+
+echo "[2/3] Build site..."
+cd site
+rm -f .env .env.local .env.secrets .env.site-configs
+rm -rf .next
+NODE_ENV=production ADMIN_URL=http://localhost:8000 npx next build
```
nsams, line 11: why do you need ADMIN_URL for build?

build-and-run-site.sh
```diff
@@ -0,0 +1,19 @@
...
+ln -sf ../.env ./
+ln -sf ../.env.local ./
+ln -sf ../.env.secrets ./
+ln -sf ../.env.site-configs ./
+echo ""
+
+echo "[3/3] Start site..."
+NODE_ENV=production SITE_PREVIEW_SECRET=local npx dotenv -e .env.site-configs -- node server.js
```
nsams, line 19: shouldn't we have the other .envs as well? For example [this](https://github.com/vivid-planet/comet-starter/blob/main/site/src/middleware.ts#L33C60-L33C76) needs API_URL_INTERNAL

build-and-run-site.sh
nsams: > which strips these headers

:thinking:

build-and-run-site.sh
johnnyomair: `.env.secrets` could be used in projects to inject server-side secrets, so I think we should include it here.

johnnyomair (approved):

johnnyomair (approved):

package.json
```diff
@@ -3,6 +3,7 @@
     "private": true,
     "scripts": {
         "create-site-configs-env": "npx @comet/cli inject-site-configs -f site-configs/site-configs.ts -i .env.site-configs.tpl -o .env.site-configs",
+        "build-and-run-site": "./build-and-run-site.sh; npx dev-pm start site",
```
johnnyomair, line 6:
```suggestion
        "build-and-run-site": "./build-and-run-site.sh; npx dev-pm restart site",
```

Should this be restart? What if I run this script, but the site wasn't running before?

build-and-run-site.sh
dkarnutsch, line 1: This script needs some sort of explanation somewhere. Without this, it might be mistaken for local development or deployed usage

nsams:
imho:
- I would remove the `dev-pm stop site` (it's logical to me that I can't have a running site and start another one)
- I would remove the `dev-pm start site` (not needed if it doesn't stop in the first place. And might potentially start site that was not running at all before)
- I would remove the package.json script (it's a bash script already, like install.sh. no need for npm here)

But that is just my opinion, feel free to ignore.

nsams (approved):

thomasdax98 (approved):

johnnyomair (approved):

dkarnutsch (approved):

PR #356, opened 2024-09-13
Introduce env APP_ENV and use it in AccessControlService

fraxachun:


johnnyomair (commented):
Alternatively, we could add a more specific `ALLOW_VIVID_USERS` variable. Is having this app env variable okay with twelve factor? It's still coming from the environment, but we have additional ifs in the code.

johnnyomair (approved):

dkarnutsch (approved):
It would be nice to display the env as a badge in the Admin (or at least in the html title)

nsams:
I also don't think we should have such a generic named env var, as it makes devs too easy to use it where a real env var would be more appopriate. (12-factor)

> A distinction between the environments will be necessary also for other cases than applying vivid users.

what cases do you think of?

thomasdax98 (approved):

PR #358, opened 2024-09-13
Allow setting base-path for admin

fraxachun:
Precondition to simplify deployment (we can eleminate the second auth-proxy then).


admin/src/config.tsx
```diff
@@ -21,6 +21,7 @@ export function createConfig() {
         ...cometConfig,
         apiUrl: environmentVariables.API_URL,
         adminUrl: environmentVariables.ADMIN_URL,
+        basename: environmentVariables.ADMIN_BASE_PATH,
```
johnnyomair, line 24: I'd you basePath here because basename could be confusing (I don't know why react-router calls it like this)

PR #362, opened 2024-09-16
Handle non-existing assets correctly

fraxachun:
Next will try to render a "normal" page when a non-existing asset (e.g. /assets/foo) is requested. This lead to misleading 500 errors as "assets" would be used as parameter for `/app/[domain]`, causing our site config setup to fail. To fix this, we add an additional page `/app/assets/[[...path]]` which always returns 404.

### DIFF
--- a/admin/server/server.js
+++ b/admin/server/server.js
@@ -8,7 +8,7 @@ const app = express();
 const port = process.env.APP_PORT ?? 3000;
 
 // Read index.html file
-let indexFile = fs.readFileSync("build/index.html", "utf8");
+let indexFile = fs.readFileSync("../build/index.html", "utf8");
 
 // Replace environment variables
 indexFile = indexFile.replace(/\$([A-Z_]+)/g, (match, p1) => {
@@ -42,7 +42,7 @@ app.get("/status/health", (req, res) => {
 });
 
 app.use(
-    express.static("./build", {
+    express.static("../build", {
         setHeaders: (res, path, stat) => {
             if (path.endsWith(".js")) {
                 // The js file is static and the index.html uses a parameter as cache buster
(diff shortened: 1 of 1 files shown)


nsams:
couldn't you instead avoid the 500 by returning null (instead of throwing Error) and call `notFound()` in page.tsx if `getSiteConfigForDomain` returned null?

site/src/util/siteConfig.ts
```diff
@@ -9,7 +9,13 @@ export function getHostByHeaders(headers: Headers) {
 }
 
 export function getSiteConfigForDomain(domain: string) {
-    const siteConfig = getSiteConfigs().find((siteConfig) => siteConfig.scope.domain === domain);
+    let siteConfig;
+    if (domain === "assets") {
```
johnnyomair, line 13: We should add a comment here explaining why this if is needed.

johnnyomair (commented):
It's odd that this needs to be done in getSiteConfigForDomain, but I can't think of a better solution. We could try to get rid of the assets folder altogether, as we discussed in https://github.com/vivid-planet/comet-starter/pull/283#discussion_r1696577469.

site/src/util/siteConfig.ts
```diff
@@ -9,7 +9,14 @@ export function getHostByHeaders(headers: Headers) {
...
+        // /assets paths they are not handled by middleware.ts, therefore no rewriting has taken place and we cannot rely on params.domain
```
johnnyomair, line 14:
```suggestion
        // /assets paths are not handled by the middleware, therefore no rewriting has taken place and we cannot rely on params.domain
```

thomasdax98 (dismissed):

johnnyomair (changes requested):
As discussed yesterday, please change this to a not found page in `src/app/assets/[[...path]]/page.tsx`

nsams (approved):

johnnyomair (approved):

johnnyomair:
Remove the recommended-change label since this should be done like in https://github.com/vivid-planet/comet-starter/pull/748 instead.

PR #364, opened 2024-09-16
Make local dev more similar to deployed instances

fraxachun:
The API always uses the static authed user, also during rendering of the site. When the site tries to render a page it calls pageTreeNodeByPath of the API alongside with the current scope. If a "wrong" scope (e.g. a language which does not exist) is sent, the API locally returns an access forbidden because the current authed user does not have access to this scope.

With this PR the API locally uses the system user for these types of requests thus return a valid response allowing the site to call notFound().

COM-681


dkarnutsch (approved):

nsams (approved):
should this also be changed in demo?

PR #367, opened 2024-09-18
Add compatibility with readonly filesystems

Vierlip:
Update server to replace environment variables in-memory

- Replace environment variables directly in-memory
- Serve modified index.html for root and fallback routes


admin/server/server.js
```diff
@@ -2,10 +2,19 @@
...
+const fs = require("fs");
 
 const app = express();
 const port = process.env.APP_PORT ?? 3000;
 
+// Read index.html file
+let indexFile = fs.readFileSync(`${__dirname}/../build/index.html`, 'utf8');
+
+// Replace environment variables
+const envsubIndexFile = indexFile.replace(/\$([A-Z_]+)/g, (match, p1) => {
```
johnnyomair, line 14: As the indexFile is mutable, we could also do `indexFile = indexFile.replace(...)`. This would remove the need for the second variable (which has a strange name IMO)

admin/server/server.js
```diff
@@ -32,8 +41,12 @@ app.get("/status/health", (req, res) => {
     res.send("OK!");
 });
 
+app.get("/", (req, res) => {
+    res.send(envsubIndexFile);
+});
```
johnnyomair, line 46: Why is this necessary? Shouldn't this be handled by the catch all [here](https://github.com/vivid-planet/comet-starter/pull/367/files#diff-cd5efe3ca968a6240f067eb4ef9db36f9b43a2639201c257c9ce0f1de0db8ac5R68)?

admin/server/server.js
```diff
@@ -53,7 +66,7 @@ app.use(
 
 // As a fallback, route everything to index.html
 app.get("*", (req, res) => {
-    res.sendFile(`index.html`, { root: `${__dirname}/../build/`, headers: { "cache-control": "no-store" } });
```
johnnyomair, line 56: I suppose the cache-control header isn't necessary anymore?

admin/server/server.js
```diff
@@ -32,8 +41,12 @@ app.get("/status/health", (req, res) => {
...
 
+app.get("/", (req, res) => {
+    res.send(envsubIndexFile);
+});
+
 app.use(
-    express.static("../build", {
+    express.static(`${__dirname}/../build`, {
         setHeaders: (res, path, stat) => {
             if (path.endsWith(".html")) {
```
johnnyomair, line 51: This won't match index.html anymore, so we should remove it

admin/server/server.js
```diff
@@ -2,10 +2,19 @@
...
 const app = express();
 const port = process.env.APP_PORT ?? 3000;
 
+// Read index.html file
+let indexFile = fs.readFileSync(`${__dirname}/../build/index.html`, 'utf8');
+
+// Replace environment variables
+indexFile = indexFile.replace(/\$([A-Z_]+)/g, (match, p1) => {
+    return process.env[p1] || "";
+});
```
nsams, line 16: you don't use the `envsub` dependency anymore, can be removed from dependencies.

Or you use it instead of the hand made implementation

admin/server/server.js
```diff
@@ -33,7 +42,7 @@ app.get("/status/health", (req, res) => {
 });
 
 app.use(
-    express.static("../build", {
+    express.static(`${__dirname}/../build`, {
```
nsams, line 45: I would not change that to __dirname, using cwd is imho preferrable

admin/server/package.json
```diff
@@ -2,12 +2,10 @@
     "name": "server",
     "private": true,
     "scripts": {
-        "preserve": "envsub --syntax dollar-basic ../build/index.html ../build/index.html",
         "serve": "node server.js"
     },
     "dependencies": {
         "compression": "^1.7.4",
-        "envsub": "^4.1.0",
```
dkarnutsch, line 10: You need to execute ./install.sh to update to package-lock.json

dkarnutsch (changes requested):

admin/server/server.js
Vierlip: Fix: 09be0bcb2e03df369054bf438043b4569df5b158

admin/server/server.js
Vierlip: Was removed here: 9316441681f16f1aefdc50219fd867ee2bf7eb9f

admin/server/server.js
Vierlip: Correct. The cache-control header is not necessary when using res.send() because you are sending the content of the indexFile directly as a string, rather than serving a static file.

admin/server/server.js
Vierlip: Was removed here: 76b0b079f63b55ba60e9d2a735be0e16b183bcf1

admin/server/server.js
Vierlip: Dependency was removed here: 6161e386053f8933c8f1aa14436c9b0eefd66705

admin/server/server.js
Vierlip: Fix: 4f70a0f003c230f0a63e156926be71971e25e695

admin/server/package.json
Vierlip: Update: 7b995445c05f5ffa0fd35e06e091155cfa45471f

johnnyomair (approved):

Schreglmann (approved):

admin/server/server.js
Vierlip: Fix: 9fce49460d079f18d23ec98c890a7e56c9b48947

johnnyomair (dismissed):

dkarnutsch (dismissed):

admin/server/server.js
```diff
@@ -2,10 +2,19 @@
 const express = require("express");
 const compression = require("compression");
 const helmet = require("helmet");
+const fs = require("fs");
 
 const app = express();
 const port = process.env.APP_PORT ?? 3000;
 
+// Read index.html file
+let indexFile = fs.readFileSync(`${process.cwd()}/build/index.html`, "utf8");
```
nsams, line 11:
```suggestion
let indexFile = fs.readFileSync("build/index.html", "utf8");
```

admin/server/server.js
```diff
@@ -33,13 +42,9 @@ app.get("/status/health", (req, res) => {
 });
 
 app.use(
-    express.static("../build", {
+    express.static(`${process.cwd()}/build`, {
```
nsams, line 45:
```suggestion
    express.static("./build", {
```

johnnyomair (approved):

nsams (approved):

PR #372, opened 2024-09-26
Revert "API: Use config for ORM config"

fraxachun:
This reverts commit e0fd283b2061c08aae29f3a8ac1f839e3d30f404.

## Description

Problem is that the comet-api-generator is executed in the pipeline where no envs are available but loads the orm-config: https://github.com/vivid-planet/comet/blob/main/packages/api/cms-api/src/generator/generate.ts#L48

After that the config gets loaded (https://github.com/vivid-planet/comet-starter/pull/365/files#diff-52cfbc7ee4148907d1c6a313cf2e25b3cb4f4accbad4352cc42d09395814cd27R6), resulting in error Messages for class-validator.

## Screenshots/screencasts

image

## Related tasks and documents

https://github.com/vivid-planet/comet-starter/pull/365


johnnyomair (approved):
It's strange that we didn't notice this in the Starter... maybe because we have no entities that are generated with the API Generator?

IMO we should still use the config in ormconfig.ts. Maybe we can find a solution where the generator works without the ORM config (why is this even needed?)

johnnyomair:
> Sure, but I could not find a possibility. Since currently the creation of a new project results in breaking pipeline I would be happy to merge this asap and postpone a solution.

Yes, I was talking about a follow-up solution (maybe also requires changes in the core).

dkarnutsch (approved):

PR #375, opened 2024-09-30
Add envs for ACL

fraxachun:
## Description

This PR adds environment variables where admins can be declared. The allows having different admins for different environments.

Note that this PR does not provide a way to give somewhat like "default permissions" (e.g. all permission but without user management). This is out of scope for this PR.

## Related tasks and documents

This PR supersedes https://github.com/vivid-planet/comet-starter/pull/356


thomasdax98:
This seems more complicated and less flexible to me than https://github.com/vivid-planet/comet-starter/pull/356

api/src/auth/access-control.service.ts
```diff
@@ -1,35 +1,32 @@
...
             return UserPermissions.allPermissions;
         }
 
-        if (user.email === staticUsers.editor.email) {
+        if (this.isUser(user)) {
             return [{ permission: "pageTree" }];
         }
 
         return [];
     }
```
thomasdax98, line 22: How would I implement something like this?

```ts
// Vivid Planet employees have default permissions in non-prod envs
if (user.isVividPlanet && this.config.appEnv !== "production") {
    return DEFAULT_PERMISSIONS;
}
```

Do I add another env variable `ACL_GIVE_VIVID_USERS_DEFAULT_PERMISSIONS`?

thomasdax98 (commented):
This seems more complicated and less flexible to me than https://github.com/vivid-planet/comet-starter/pull/356

api/src/auth/access-control.service.ts
```diff
@@ -38,4 +35,12 @@ export class AccessControlService extends AbstractAccessControlService {
...
+
+    private isAdmin(user: User) {
+        return this.config.acl.adminEmails.includes(user.email) || this.config.acl.adminEmailDomains.includes(user.email.split("@")[1]);
+    }
+
+    private isUser(user: User) {
```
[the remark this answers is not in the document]
johnnyomair: Fine by me. We can also keep the original naming, if it doesn't cause confusion among our colleagues.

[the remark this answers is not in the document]
johnnyomair: Admin vs. superuser has always been confusing to me, and it seems that superuser is more powerful than admin: https://stackoverflow.com/questions/20469174/whats-the-difference-between-staff-admin-superuser-in-django

So maybe using one of the two previous suggestions is better.

.env
```diff
@@ -39,6 +39,10 @@ API_URL=http://${DEV_DOMAIN:-localhost}:${API_PORT}/api
 API_URL_INTERNAL=http://localhost:$API_PORT/api
 CORS_ALLOWED_ORIGIN="^http:\/\/(localhost|.*\.dev\.vivid-planet\.cloud|192\.168\.\d{1,3}\.\d{1,3}):\d{2,4}"
 BASIC_AUTH_SYSTEM_USER_PASSWORD=password
+ACL_ADMIN_EMAILS=admin@customer.com
+ACL_ADMIN_EMAIL_DOMAINS=vivid-planet.com
+ACL_SUPERUSER_EMAILS=german-editor@customer.com
+ACL_SUPERUSER_EMAIL_DOMAINS=
```
thomasdax98, line 45: ACL_ALL_PERMISSIONS_EMAILS=
ACL_ALL_PERMISSIONS_DOMAINS=

johnnyomair (commented):
Please also provided the basis for a `projectLeads` array defined in the code.

johnnyomair:
> The projectLeads also have to be set via envs.

I thought we decided on adding this in code?

johnnyomair:
> Adding as env still is in code (at least in the way we do it).

Fine by me. @thomasdax98 @nsams @dkarnutsch do you agree?

nsams:
> Adding as env still is in code (at least in the way we do it).

I also thought about splitting this into two parts:

- a list of emails hardcoded in `access-control.service.ts`, containing the lead-devs of the project (will be the same for all deployments)
- a list of emails injected using env var, containing the infra-team (will change over time when the infra-team changes)

Of course the we can put both into a single env, but to have an opinion about that I would like to see the relevant deployment config. (Don't know if you can post that here)

api/src/auth/access-control.service.ts
```diff
@@ -3,6 +3,8 @@ import { Inject, Injectable } from "@nestjs/common";
 import { Config } from "@src/config/config";
 import { CONFIG } from "@src/config/config.module";
 
+const allPermissionEmails: string[] = [];
```
johnnyomair, line 6: This should probably be allPermission**s**Emails. Could this be confusing with ` this.config.acl.allPermissionsEmails`?

johnnyomair: `projectLeads` 😁 But I'm okay with this as well

johnnyomair (approved):

dkarnutsch (approved):

nsams (approved):

thomasdax98 (approved):

PR #382, opened 2024-10-07
Hotfix Admin Server

dkarnutsch:
- path of `index.html` is invalid (path is relative to `admin/server/server.js`)
- `res.send` signature is invalid (it tries to send the content as HTTP status code)

introduced by #367


nsams (approved):

nsams:
@Vierlip it is your responsibility to apply (and test) suggestions

johnnyomair (approved):

PR #384, opened 2024-10-08
Refactor Typography

SebiVPS:
## Description

Refactor typography component to export only the styled component instead of the react function component.
This allows to use the component with other styled components.

This can be undone: https://github.com/vivid-planet/comet-starter/pull/282/commits/43aa2675fd4808b425a5c1b0bc3fd1e86a46be6c

### DIFF
--- a/site/src/common/components/Typography.tsx
+++ b/site/src/common/components/Typography.tsx
@@ -1,4 +1,4 @@
-import { HTMLAttributes, PropsWithChildren } from "react";
+import { ComponentProps } from "react";
 import styled, { css } from "styled-components";
 
 export type TypographyVariant = "h600" | "h550" | "h500" | "h450" | "h400" | "h350" | "p300" | "p200";
@@ -180,7 +180,7 @@ const typographyVariantStyle: Record<TypographyVariant, ReturnType<typeof css>> 
     `,
 };
 
-const variantToElementMap: Record<TypographyVariant, keyof HTMLElementTagNameMap> = {
+const variantToElementMap: Record<TypographyVariant, "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p"> = {
     h600: "h1",
     h550: "h2",
     h500: "h3",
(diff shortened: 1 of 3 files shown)


site/src/common/components/Typography.tsx
```diff
@@ -191,30 +191,17 @@ const variantToElementMap: Record<TypographyVariant, keyof HTMLElementTagNameMap
     p200: "p",
 };
 
-export interface TypographyProps extends PropsWithChildren<HTMLAttributes<HTMLElement>> {
-    component?: keyof HTMLElementTagNameMap;
+export const Typography = styled.div.attrs<{
+    as?: unknown;
```
SebiVPS, line 195: To avoid private import, i typed it here as unkown.

`import { KnownTarget } from "styled-components/dist/types";`

johnnyomair: Have you tried if type inference still works when doing something like `<Typography as={LinkBlock} data={/* ... */} />`?

SebiVPS: Yes 👍  works

johnnyomair (approved):

site/src/common/components/Typography.tsx
```diff
@@ -223,3 +210,5 @@ const Text = styled.div<TextProps>`
             }
         `};
 `;
+
+export type TypographyProps = ComponentProps<typeof Typography>;
```
jamesricky, line 214: I don't think we actually need this 🤷🏼‍♂️ 

```suggestion
```

jamesricky (approved):

site/src/common/components/Typography.tsx
SebiVPS: The props are used here:
https://github.com/vivid-planet/comet-starter/blob/88486189a73bcb1d4322e41855259d33806c220f/site/src/common/blocks/RichTextBlock.tsx#L13

PR #391, opened 2024-10-10
Remove `mikro-orm:drop` script

johnnyomair:
The script doesn't work because MikroORM doesn't know about entities defined in the library.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -18,7 +18,6 @@
         "lint:generated-files-not-modified": "npm run api-generator && git diff --exit-code HEAD --",
         "lint:tsc": "tsc --project ./tsconfig.lint.json",
         "mikro-orm": "mikro-orm",
-        "mikro-orm:drop": "mikro-orm schema:drop -r",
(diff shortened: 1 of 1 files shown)


api/package.json
```diff
@@ -18,7 +18,6 @@
         "lint:generated-files-not-modified": "npm run api-generator && git diff --exit-code HEAD --",
         "lint:tsc": "tsc --project ./tsconfig.lint.json",
         "mikro-orm": "mikro-orm",
```
johnnyomair, line 20: IMO we could remove this script as well. I doubt anyone uses it...

api/package.json
```diff
@@ -18,7 +18,6 @@
...
-        "mikro-orm:drop": "mikro-orm schema:drop -r",
         "mikro-orm:migration:generate": "mikro-orm migration:create",
```
johnnyomair, line 21: Also, why do we use the term generate, and not create?

nsams (approved):

dkarnutsch (approved):

thomasdax98 (approved):

PR #388, opened 2024-10-09
Docker-Compose Deployment, #588 

dkarnutsch:
#588 

## Open TODOs/questions

-   [x] TBD: Move site-config.d.ts to subfolder to enable mounting

PR #401, opened 2024-10-17
Move site-configs.ts to site-configs directory

dkarnutsch:
## Description

As a prerequisite for #388, I need to move `site-configs.ts` to a directory in order to mount it during build. Additionally, I think this improves readability.


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "starter",
     "private": true,
     "scripts": {
-        "create-site-configs-env": "npx @comet/cli inject-site-configs -i .env.site-configs.tpl -o .env.site-configs",
+        "create-site-configs-env": "npx @comet/cli inject-site-configs -f ../../../../../site-configs/site-configs.ts -i .env.site-configs.tpl -o .env.site-configs",
```
dkarnutsch, line 5: This is due to https://github.com/vivid-planet/comet/blob/e2eb8265316f76980c0f47cb2ccdce2106ea7847/packages/cli/src/commands/site-configs.ts#L16.

It should always prepend `process.cwd()`. This should be fixed in Comet.

site-configs/site-configs.ts
```diff
@@ -15,9 +15,9 @@ const getSiteConfigs = async (env: string): Promise<SiteConfig[]> => {
         throw new Error(`Invalid environment: ${env}`);
     }
     
-    const path = `${__dirname}/site-configs`;
+    const path = `${process.cwd()}/site-configs`;
 
-    const files = (await fs.readdir(path)).filter((file) => !file.startsWith("_"));
+    const files = (await fs.readdir(path)).filter((file) => (!file.startsWith("_") && !["site-configs.d.ts", "site-configs.ts"].includes(file)));
```
dkarnutsch, line 20: I don't like this solution. I think we should use a more declarative solution instead of relying on some magic.

I would propose to introduce a variable containing all sites. This was briefly discusses in one JFX.

johnnyomair (commented):
We should implement both your suggestions before merging this.

site-configs/site-configs.ts
nsams: Or use a standard, non-dynamic import, that would also improve typing (as you don't need `as ... GetSiteConfig`)

dkarnutsch: cc4fc17e1de80eccad0016a26c273bb6d90b4fb1

package.json
johnnyomair: @dkarnutsch do you want to merge it like this and change it once the fix is released in core?

dkarnutsch: yes 👍

johnnyomair (approved):

nsams (approved):

thomasdax98 (approved):

PR #404, opened 2024-10-17
Fix dotenv files hierarchy

johnnyomair:
If a environment variable was already defined in `.env`, a different value in `.env.secrets` or `.env.local` wouldn't overwrite the value. Changing the order of the dotenv files resolves the issue.


admin/package.json
```diff
@@ -4,7 +4,7 @@
     "scripts": {
         "postinstall": "cd server && npm install",
         "serve": "cd server && npm run serve",
-        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -e .env -e .env.local -e .env.secrets -e .env.site-configs -- vite",
+        "start": "npm run intl:compile && run-p gql:types generate-block-types && dotenv -e .env.secrets -e .env.local -e .env -e .env.site-configs -c -- vite",
```
nsams, line 7: is the -c added intentionally?

johnnyomair: Nope, `git add .` 🤦🏼‍♂️ Fixed: 8ff9f8b604f700369532690bb160a9b9ae190ce3

dkarnutsch (approved):

nsams (approved):

PR #408, opened 2024-10-23
Implement a REPL for the API

abel-castro:
Implement a [REPL](https://nodejs.org/en/learn/command-line/how-to-use-the-nodejs-repl) (Read Evaluate Print Loop). The REPL will be useful for example for testing MikroORM queries that interact with our real development database. 

Start it with: 
```sh
npm run --prefix api repl
```
image

Based on the [NestJS REPL recipe ](https://docs.nestjs.com/recipes/repl).

### DIFF
--- a/api/src/repl.ts
+++ b/api/src/repl.ts
@@ -1,13 +1,18 @@
 import "tsconfig-paths/register";
 
 import { MikroORM } from "@mikro-orm/core";
+import { NestFactory } from "@nestjs/core";
 import { repl } from "@nestjs/core/repl";
+import { NestExpressApplication } from "@nestjs/platform-express";
 
 import { AppModule } from "./app.module";
-import { ormConfig } from "./db/ormconfig";
+import { createConfig } from "./config/config";
 
 async function bootstrap() {
-    const replServer = await repl(AppModule);
+    const config = createConfig(process.env);
+    const appModule = AppModule.forRoot(config);
+    const app = await NestFactory.create<NestExpressApplication>(appModule);
+    const replServer = await repl(appModule);
 
     // preserve history between refreshes of the repl
     replServer.setupHistory(".nestjs_repl_history", (err) => {
@@ -17,13 +22,7 @@ async function bootstrap() {
     });
 
     // MikroORM setup
-    const orm = await MikroORM.init({
-        ...ormConfig,
-        debug: true,
-        allowGlobalContext: true,
-        entities: ["./dist/**/entities/*entity.js"],
-        entitiesTs: ["/**/entities/*.entity.ts"],
-    });
+    const orm = app.get(MikroORM);
 
     // Context shortcuts
     const replContext = replServer.context;
--- a/api/src/repl.ts
+++ b/api/src/repl.ts
@@ -9,6 +9,10 @@ import { AppModule } from "./app.module";
 import { createConfig } from "./config/config";
 
 async function bootstrap() {
+    if (process.env.NODE_ENV === "production") {
+        throw new Error("Can't use REPL in production");
+    }
+
     const config = createConfig(process.env);
     const appModule = AppModule.forRoot(config);
     const app = await NestFactory.create<NestExpressApplication>(appModule);
--- a/api/src/repl.ts
+++ b/api/src/repl.ts
@@ -21,7 +21,7 @@ async function bootstrap() {
         }
     });
 
-    // MikroORM setup
+    // Get the MikroORM instance from the app
     const orm = app.get(MikroORM);
 
     // Context shortcuts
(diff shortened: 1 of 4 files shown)


api/src/repl.ts
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    // MikroORM setup
+    const orm = await MikroORM.init({
+        ...ormConfig,
+        debug: true,
+        allowGlobalContext: true,
+        entities: ["./dist/**/entities/*entity.js"],
+        entitiesTs: ["/**/entities/*.entity.ts"],
+    });
```
nsams, line 26: can't you use the existing mikro-orm instance from the MikroOrmModule defined in AppModule?

abel-castro: Honestly, I don’t know how to do that, nor the benefits of doing so 😅. Could you maybe explain me? 
My idea was to do it this way to be able to override the default Mikro-ORM config values, such as `debug` or `allowGlobalContext`, as this could be quite useful while developing.

johnnyomair (dismissed):

api/src/repl.ts
thomasdax98, line 26: This doesn't seem to work correctly since we pass the unconfigured AppModule. (It probably only works for the DB, but not for the actual Nest App):
Bildschirmfoto 2024-10-24 um 14 49 07

As you can see, the AppModule has no services.

---

I guess we should pass the configured appModule like this:

```suggestion
    const config = createConfig(process.env);
    const appModule = AppModule.forRoot(config);
    const app = await NestFactory.create<NestExpressApplication>(appModule);

    const replServer = await repl(appModule);

    // preserve history between refreshes of the repl
    replServer.setupHistory(".nestjs_repl_history", (err) => {
        if (err) {
            console.error(err);
        }
    });

    // MikroORM setup
    const orm = app.get(MikroORM);
```

Then I get an actual list of the services and can access them:

Bildschirmfoto 2024-10-24 um 14 49 37


This also allows getting the MikroORM instance from the app as @nsams suggested:

```ts
const orm = app.get(MikroORM);
```

thomasdax98 (commented):
That's a really cool feature for debugging and development 🚀

api/src/repl.ts
abel-castro: You are totally right. I have added your suggestions in the commit https://github.com/vivid-planet/comet-starter/pull/408/commits/2da63ec296655e15c0eafe6bcd8a8e0adb8280c2

johnnyomair (dismissed):

api/src/repl.ts
nsams: > nor the benefits of it

testing mikro-orm queries ist just one example of what you can do with this REPL. You can also call any other service/resolver/controller....

api/src/repl.ts
```diff
@@ -0,0 +1,32 @@
...
+        }
+    });
+
+    // MikroORM setup
+    const orm = app.get(MikroORM);
+
+    // Context shortcuts
+    const replContext = replServer.context;
+    replContext.orm = orm;
+    replContext.em = orm.em;
```
nsams, line 30: strictly speaking you are not setting up MikroORM anymore

```suggestion
    // Context shortcuts
    const replContext = replServer.context;
    replContext.orm = app.get(MikroORM);
    replContext.em = app.get(MikroORM).em;
```

abel-castro: I agree with you and corrected that in the commit https://github.com/vivid-planet/comet-starter/pull/408/commits/a11ea948a6c926dc44a61ae85e8b790d47cd8137

api/src/repl.ts
```diff
@@ -0,0 +1,32 @@
...
+import { MikroORM } from "@mikro-orm/core";
+import { NestFactory } from "@nestjs/core";
+import { repl } from "@nestjs/core/repl";
+import { NestExpressApplication } from "@nestjs/platform-express";
+
+import { AppModule } from "./app.module";
+import { createConfig } from "./config/config";
+
+async function bootstrap() {
+    const config = createConfig(process.env);
```
thomasdax98, line 12: @abel-castro We decided that we only want to support REPL for local debugging. We don't want to use it in deployed envs as there is a risk of tampering with data without a log. This contradicts our security guidelines.

Can you please check here if you are in a deployed environment and throw an error. Should look similar to this:

```ts
if (process.env.NODE_ENV === "production") {
    throw new Error("Can't use REPL in production");
}
```

abel-castro: I added that in commit https://github.com/vivid-planet/comet-starter/pull/408/commits/2191d8c168fafb4e66a79c0572b41429c4db5da3. The output will be something like this: 

image

dkarnutsch (approved):

nsams (approved):

thomasdax98 (approved):

PR #410, opened 2024-10-24
RFC: Implement Button with CSS Modules

johnnyomair:
We want to replace  styled-components with another styling solution to support Server Components. This PR demonstrates how the Button component, which has dynamic styles, could be implemented using CSS Modules.

Advantages of CSS Modules over CSS-in-JS
- Reduced Bundle Size
- Works without JS
- Works with Server Components out-of-the-box
- No runtime required

Disadvantages
- Inferior Dev UX due to missing types (maybe this could be fixed)
- Dynamic styles are more complicated

This PR is just to start a discussion

### DIFF
--- a/site/src/app/global.css
+++ b/site/src/app/global.css
@@ -1,0 +1,14 @@
+:root {
+    --palette-primary-main: #1e88e5;
+    --palette-primary-dark: #155fa0;
+    --palette-primary-contrast-text: #ffffff;
+
+    --palette-gray-50: #fafafa;
+    --palette-gray-200: #eeeeee;
+    --palette-gray-400: #bdbdbd;
+
+    --spacing-s400: 16px;
+    --spacing-s500: 24px;
+
+    --font-family: Arial, sans-serif;
+}
--- a/site/src/common/components/Button.module.css
+++ b/site/src/common/components/Button.module.css
@@ -1,0 +1,67 @@
+.root {
+    display: inline-flex;
+    padding: var(--spacing-s400) var(--spacing-s500);
+    border-radius: 4px;
+    cursor: pointer;
+    justify-content: center;
+    align-items: center;
+    transition: background-color 0.2s ease-out, color 0.2s ease-out, border-color 0.2s ease-out;
+
+    text-align: center;
+    text-decoration: none;
+    font-family: var(--font-family);
+    font-size: 16px;
+    font-weight: 700;
+    line-height: 110%;
+
+    &:disabled {
+        pointer-events: none;
+    }
+}
+
+.contained {
+    background-color: var(--palette-primary-main);
+    color: var(--palette-primary-contrast-text);
+    border: 1px solid var(--palette-primary-main);
+
+    &:hover {
+        background-color: var(--palette-primary-dark);
+        border-color: var(--palette-primary-dark);
+    }
+
+    &:disabled {
+        background-color: var(--palette-gray-50);
+        color: var(--palette-gray-400);
+        border-color: var(--palette-gray-200);
+    }
+}
+
+.outlined {
+    background-color: transparent;
+    color: var(--palette-primary-main);
+    border: 1px solid var(--palette-primary-main);
+
+    &:hover {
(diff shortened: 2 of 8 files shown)


site/src/common/components/Button.module.css
```diff
@@ -0,0 +1,67 @@
...
+        border-color: var(--palette-gray-200);
+    }
+}
+
+.outlined {
+    background-color: transparent;
+    color: var(--palette-primary-main);
+    border: 1px solid var(--palette-primary-main);
+
+    &:hover {
```
nsams, line 44: who or what does make this working? Or is this valid css in 2024?

johnnyomair: Yes: https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector

However, it doesn't meet our support requirements on Can I use... to be honest, I haven't thought about this too much.

nsams:
Interesting would also be an example where a component has multiple elements, not just a root

site/src/common/components/Button.module.css
nsams: we could add postcss....

is something like autoprefixer needed nowadays? how does styled-components handle this?

johnnyomair: > we could add postcss....

Yes, if we decide to switch, we should invest some time into improving the developer experience.

> is something like autoprefixer needed nowadays? how does styled-components handle this?

I don't know. styled-components does it out-of-the-box: https://styled-components.com/docs/basics#motivation

johnnyomair:
> Interesting would also be an example where a component has multiple elements, not just a root

Here you go: 98158fd34f9bf31e3f38851eda5a69a17b92899e

site/src/documents/pages/blocks/TeaserItemBlock.module.css
```diff
@@ -0,0 +1,62 @@
...
+        display: none;
+    }
+
+    .mediaDesktop {
+        display: block;
+        flex: 1;
+    }
+}
+
+@media (min-width: 900px) {
```
nsams, line 56: I guess there is not way to avoid hardcoding the breakpoints

johnnyomair: I tried using CSS variables, but that didn't work

site/src/common/components/Button.module.css
jamesricky: Importing `scss` files instead of `css` files seems to work out of the box, this would allow nested selectors independent of browser support.

site/src/app/global.css
```diff
@@ -0,0 +1,18 @@
+:root {
+    --palette-primary-main: #1e88e5;
+    --palette-primary-dark: #155fa0;
+    --palette-primary-contrast-text: #ffffff;
+
+    --palette-gray-50: #fafafa;
+    --palette-gray-200: #eeeeee;
+    --palette-gray-300: #e0e0e0;
+    --palette-gray-400: #bdbdbd;
```
jamesricky, line 9: I'd omit the `palette-` prefix, `--primary-main` or `--gray-50` es precise enough and easier to write imo.

jamesricky (commented):
I like this 👍🏼

site/src/common/components/Button.module.css
johnnyomair: Switched to Sass: 78a44b48b0353128976efd91c60d9c91f8b535ed

site/src/documents/pages/blocks/TeaserItemBlock.module.css
johnnyomair: I used Sass variables: e347a2b50b61f44d3e85cb2434520450c802fc03

However, the need to import the partial Sass file isn't great.

johnnyomair:
> Inferior developer experience due to missing types (maybe this could be fixed)

Added types with a small util here: 5378f9d22cbbc6602875e40ae5fdbc3e458c4f0bs

image

image

PR #415, opened 2024-11-04
Move CSP-Headers to middleware

fraxachun:
## Description

Until this PR we defined the Content-Security-Policy header in next.config.js, which is then created during build. The problem is that (at least in our pipelines) environment variables (like process.env.ADMIN_URL) are not available during the build resulting in empty header values.

This has not turned out as a problem until now, because when using the site embedded in an I-Frame it was protected by the OAuth2-Proxy which strips this header. But since https://github.com/vivid-planet/comet/pull/2554 this will produce an error.

The solution is to send this header in the middleware where we have access to the environment variables. However, this (probably) comes with a speed penalty since the code is executed for every request.


thomasdax98 (dismissed):

nsams:
Not ideal, I would like to have the middleware as small as possible. But there seems to be no other way to do this.

Maybe, you could extract that to an function (and file)?

johnnyomair (commented):
What about  the block preview, which isn't handled by the middleware? Shouldn't we add the headers for it as well? And redirects/rewrites (although I believe we don't need them there)

nsams (dismissed):

nsams:
I will come up with a proposal for more elegantly combining multiple middlewares (in a future PR)

max-debug022 (approved):

nsams:
> I will come up with a proposal for more elegantly combining multiple middlewares (in a future PR)

https://github.com/vivid-planet/comet/pull/2728

johnnyomair (approved):

thomasdax98 (dismissed):

site/next.config.js
```diff
@@ -24,7 +24,7 @@ const nextConfig = {
     experimental: {
         optimizePackageImports: ["@comet/cms-site"],
     },
-    // https://nextjs.org/docs/advanced-features/security-headers
+    // https://nextjs.org/docs/advanced-features/security-headers (Content-Security-Policy is set in middleware.ts)
```
dkarnutsch, line 27:
```suggestion
    // https://nextjs.org/docs/advanced-features/security-headers (Content-Security-Policy and CORS are set in middleware.ts)
```

dkarnutsch (dismissed):

johnnyomair (approved):

PR #416, opened 2024-11-07
AccordionBlock: use height transition instead of css only solution

SebiVPS:
## Description

Because the css solution with `margin-top: -100%;` could potentially made some problems we fall back to the default ref height transition animation.


nsams:
:-1: for a non-css-only solution in starter

site/src/common/blocks/AccordionItemBlock.tsx
```diff
@@ -83,21 +97,17 @@ const AnimatedChevron = styled(SvgUse)<{ $isExpanded: boolean }>`
...
-            opacity: 1;
-            transition: margin-top 0.5s ease-out, opacity 0.3s linear 0.4s;
-        `}
+const ContentWrapperInner = styled.div`
+    padding: ${({ theme }) => theme.spacing.S500} 0;
+    border-top: 1px solid ${({ theme }) => theme.palette.gray["300"]};
+
+    > *:last-child {
+        margin-bottom: 0;
+    }
```
johnnyomair, line 112: Why is this needed?

site/src/common/blocks/AccordionItemBlock.tsx
```diff
@@ -83,21 +97,17 @@ const AnimatedChevron = styled(SvgUse)<{ $isExpanded: boolean }>`
     transition: transform 0.4s ease;
 `;
 
-const ContentWrapper = styled.div`
+const ContentWrapper = styled.div<{ $isExpanded: boolean; $height: number }>`
     overflow: hidden;
+    height: ${({ $height }) => $height}px;
+    transition: height 300ms, margin-bottom 300ms;
```
johnnyomair, line 103: Why do you need to animate margin-bottom?

site/src/common/blocks/AccordionItemBlock.tsx
```diff
@@ -83,21 +97,17 @@ const AnimatedChevron = styled(SvgUse)<{ $isExpanded: boolean }>`
     transition: transform 0.4s ease;
 `;
 
-const ContentWrapper = styled.div`
+const ContentWrapper = styled.div<{ $isExpanded: boolean; $height: number }>`
```
johnnyomair, line 100: $expanded isn't used

johnnyomair:
IMO a toggle between `display: none` and `display: block` would be acceptable as well 🤷🏼‍♂️

site/src/common/blocks/AccordionItemBlock.tsx
SebiVPS: To remove the margin-bottom of the last element (block) in the content div. A padding is added by the container itself, so for example a typography would add  a default margin bottom and with the padding the space would be too much. It looks better, when the margin-bottom of the last child element is removed.

johnnyomair: Shouldn't the rich text block remove this margin? It's odd that the accordion block has to take care of this.

SebiVPS: How should he know that?

SebiVPS: IMO its better, that the accordion block handles the inset, he is in charge of all of the insets, so the bottom should be no exception.

johnnyomair: @jamesricky @mariokemetinger what do you think?

johnnyomair (commented):
Waiting for a decision here: https://github.com/vivid-planet/comet-starter/pull/416#discussion_r1839904346

site/src/common/blocks/AccordionItemBlock.tsx
jamesricky: > Shouldn't the rich text block remove this margin? It's odd that the accordion block has to take care of this.

I agree

> How should he know that?

I think we discussed a setting for `disableLastItemMargin` or something for the `RichTextBlock` 🤔 
Was this ever added? Might be the right solution for that here .

jamesricky: Ah, you can set `disableLastBottomSpacing` on `<RichTextBlock />` in the `supportedBlocks` of `AccordionItemBlock`.

mariokemetinger: @SebiVPS but what are you doing when the block has a padding instead of margin? I think this should be handled by the blocks in accordionContent itself...

jamesricky:
> 👎 for a non-css-only solution in starter

I somewhat agree, but I don't know of another reliable way of achieving this. 
And since we likely need this is 99% of projects I'd prefer this over just showing/hiding with `display: none/block`.

site/src/common/blocks/AccordionItemBlock.tsx
jamesricky: As discussed, please remove the `margin-bottom: 0;` in this PR and adapt the `RichTextBlock` in a separate PR to always remove the last paragraph's `margin-bottom`.

SebiVPS: removed 👍

site/src/common/blocks/AccordionItemBlock.tsx
```diff
@@ -83,21 +97,17 @@ const AnimatedChevron = styled(SvgUse)<{ $isExpanded: boolean }>`
...
+const ContentWrapper = styled.div<{ $height: number }>`
     overflow: hidden;
+    height: ${({ $height }) => $height}px;
+    transition: height 300ms;
 `;
 
-const ContentWrapperInner = styled.div<{ $isExpanded: boolean }>`
-    padding-bottom: ${({ theme }) => theme.spacing.S300};
-    margin-top: -100%;
-    opacity: 0;
```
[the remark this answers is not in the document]
SebiVPS: re-added the opacity animation

jamesricky (dismissed):

site/src/common/blocks/AccordionItemBlock.tsx
SebiVPS: @johnnyomair  https://github.com/vivid-planet/comet-starter/pull/416/commits/d4b22088b089fbd1bb6e8ac04baf8292d5de1aae

site/src/common/blocks/AccordionItemBlock.tsx
```diff
@@ -97,17 +97,21 @@ const AnimatedChevron = styled(SvgUse)<{ $isExpanded: boolean }>`
...
     height: ${({ $height }) => $height}px;
-    transition: height 300ms, margin-bottom 300ms;
+    opacity: 0;
+    transition: height 0.2s ease-out 0.2s, opacity 0.3s linear;
+
+    ${({ $isExpanded }) =>
+        $isExpanded &&
+        css`
+            opacity: 1;
+            transition: height 0.2s ease-out, opacity 0.3s linear 0.2s;
```
johnnyomair, line 110: Is this necessary? It seems to me that changing the transition is a mistake

site/src/common/blocks/AccordionItemBlock.tsx
```diff
@@ -44,8 +58,8 @@ export const AccordionItemBlock = withPreview(
                         <AnimatedChevron href="/assets/icons/chevron-down.svg#chevron-down" $isExpanded={isExpanded} />
                     </IconWrapper>
                 </TitleWrapper>
-                <ContentWrapper aria-hidden={!isExpanded}>
-                    <ContentWrapperInner $isExpanded={isExpanded}>
+                <ContentWrapper $height={elementHeight} $isExpanded={isExpanded}>
```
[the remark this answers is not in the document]
SebiVPS: no, that was a mistake. Nice catch 🙏

jamesricky (approved):

johnnyomair (dismissed):

mariokemetinger (approved):

nsams:
> I don't know of another reliable way of achieving this.

Setting the inline height has so many disadvantages (and bugs) - we really must give our best to avoid that.

See here for possible solutions:
https://stackoverflow.com/questions/3508605/how-can-i-transition-height-0-to-height-auto-using-css

The max-height idea seems pretty clever.

SebiVPS:
> > I don't know of another reliable way of achieving this.
> 
> Setting the inline height has so many disadvantages (and bugs) - we really must give our best to avoid that.
> 
> See here for possible solutions: https://stackoverflow.com/questions/3508605/how-can-i-transition-height-0-to-height-auto-using-css
> 
> The max-height idea seems pretty clever.

> Setting the inline height has so many disadvantages (and bugs) - we really must give our best to avoid that.
@nsams can you explain this? I do not know these.

> The max-height idea seems pretty clever.
The max-height solution has other problems (trade offs). The transition is always towards the max-height (smaller heights are not animated properly). What do we set for max-height - any content can be inserted into the accordion item?

I am no css expert, when @mariokemetinger, @jamesricky or @johnnyomair have any other, more solid, solution feel free to talk to me :)

nsams:
> can you explain this? I do not know these.

if the content changes it's height (other than window resize) the inline height won't be updated

nsams:
> The max-height solution has other problems

right. The stackoverflow question has 46 answers, maybe some other is working for us.

jamesricky:
> > The max-height idea seems pretty clever.
> 
> The max-height solution has other problems (trade offs). The transition is always towards the max-height (smaller heights are not animated properly). What do we set for max-height - any content can be inserted into the accordion item?
> 
> I am no css expert, when @mariokemetinger, @jamesricky or @johnnyomair have any other, more solid, solution feel free to talk to me :)

Animating height using the `max-height` method is a bad solution imo. 
The animation speed differs depending of the size of the content, easing does't work properly and this causes a limit to the height of the content.

SebiVPS:
> if the content changes it's height (other than window resize) the inline height won't be updated

@nsams correct me if i'm wrong, but i think there is no such case in the starter, where the content can change it's height. I think for now, this solution is working for us/the starter, better than the current one. (also most used in our projects).
If you want, @thomasdax98 can make an analyze task to find a more robust solution also respecting height changes of the content?

nsams:
> correct me if i'm wrong, but i think there is no such case in the starter, where the content can change it's height.

This is the starter, that starts anything you can't imagine. That's why I say it's an Okish solution for an project where you can argue that no content can change it's height. But not for starter.

Have you looked at the other 45 answers?

An alternative solution might be to unset the inline style after the animation completed. I don't know how easy that is nowadays.

johnnyomair:
Superseded by https://github.com/vivid-planet/comet-starter/pull/735.

PR #513, opened 2024-11-13
Site: Move resolving of redirects from middleware to ...path page

nsams:
This has better performance as we don't have to load all redirects into memory.

Demo implementation: https://github.com/vivid-planet/comet/pull/2638

The /admin redirect in #2638 is broken (when deployed), leave that in middleware to be able to access env var


site/src/app/[domain]/[language]/[[...path]]/page.tsx
```diff
@@ -1,33 +1,71 @@
...
 
-export default async function Page({ params: { path, domain, language } }: { params: { path: string[]; domain: string; language: string } }) {
-    const scope = { domain, language };
-
+export default async function Page({ params }: { params: { path: string[]; domain: string; language: string } }) {
     const { previewData } = (await previewParams()) || { previewData: undefined };
     const graphqlFetch = createGraphQLFetch(previewData);
+    const siteConfig = getSiteConfigForDomain(params.domain);
+
+    const skipPage = !siteConfig.scope.languages.includes(params.language);
```
Comment (johnnyomair) [inline site/src/app/[domain]/[language]/[[...path]]/page.tsx:34]: Should we add a comment explaining this? Or do you think it's obvious enough?

Comment (nsams) [reply site/src/app/[domain]/[language]/[[...path]]/page.tsx:34]: yes, this is not very obvious

maybe something like that?
```suggestion
    const skipPage = !siteConfig.scope.languages.includes(params.language); // redirects are not language scoped, language is invalid we still need to load them (and skip pages)
```

thomasdax98 (approved):

site/src/app/[domain]/[language]/[[...path]]/page.tsx
Comment (nsams) [reply site/src/app/[domain]/[language]/[[...path]]/page.tsx:34]:
```suggestion
    // Redirects are scoped by domain only, not by language.
    // If the language param isn't a valid language, it may still be the first segment of a redirect source.
    // In that case we skip resolving page and only check if the path is a redirect source.
    const skipPage = !siteConfig.scope.languages.includes(params.language);
```

johnnyomair (dismissed):

johnnyomair (changes requested):
Please resolve conflicts

site/src/app/[domain]/[language]/[[...path]]/page.tsx
```diff
@@ -1,33 +1,74 @@
...
     if (!data.pageTreeNodeByPath?.documentType) {
+        if (data.redirectBySource?.target) {
+            const target = data.redirectBySource?.target as RedirectsLinkBlockData;
+            let destination: string | undefined;
+            if (target.block !== undefined) {
+                switch (target.block.type) {
+                    case "internal": {
+                        const internalLink = target.block.props as InternalLinkBlockData;
+                        if (internalLink.targetPage) {
+                            destination = `${(internalLink.targetPage.scope as GQLPageTreeNodeScope).language}/${internalLink.targetPage.path}`;
```
Comment (nsams) [inline site/src/app/[domain]/[language]/[[...path]]/page.tsx:59]:
```suggestion
                            destination = `/${(internalLink.targetPage.scope as GQLPageTreeNodeScope).language}/${internalLink.targetPage.path}`;
```

johnnyomair (approved):

thomasdax98 (approved):

PR #536, opened 2024-11-19
Site: Set correct document language

johnnyomair:
We need to set the `lang` attribute for the `<html>` for certain features (e.g., `hyphens: auto` in CSS) to work correctly. For this to work, I moved the `<html>` and `<body>` tags from the root layout to the layouts where we know the language (`app/[domain]/[language]/layout.tsx`, `app/block-preview/[domain]/[language]/layout.tsx`). Doing so caused errors in outer layouts that rendered tags (e.g., `<noscript>`), therefore I have removed these layouts and moved the functionality into the inner layouts. 

**IMO we should consider the two layouts for the site and block preview as the new root layouts.**

Additionally, since the actual root layout no longer renders the `<html>` and `<body>` tags, the default not found page didn't work any longer. To fix this, I've added a custom not found page that renders the tags. We probably need to do the same for the global error page, but this is out-of-scope of this PR.


johnnyomair:
Alternatives I've thought of:
1. Set the document language in the IntlProvider using useEffect. Downside: The attribute is only set on the client
2. Detect the language in the middleware and pass a header to the root layout. Downside: Dynamic rendering

So I think this is the only viable solution.

manuelblum (approved):

nsams (approved):

thomasdax98 (approved):

dkarnutsch (approved):

PR #546, opened 2024-11-20
Activate Renovate lock file maintenance

manuelblum:
## Description
Activate [lock file maintenance](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance)  to update transient dependencies in lock file.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -30,12 +30,7 @@
         },
         {
             "groupName": "Mui",
-            "matchPackageNames": [
-                "/@mui/material/",
-                "/@mui/system/",
-                "/@mui/utils/",
-                "/@mui/lab/"
-            ]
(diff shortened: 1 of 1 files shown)


johnnyomair (commented):
> When Renovate performs lockFileMaintenance it deletes the lock file and runs the relevant package manager. That package manager creates a new lock file, where all dependency versions are updated to the latest version. Renovate then commits that lock file to the update branch and creates the lock file update PR.

Is this really a good idea?

manuelblum:
@johnnyomair I am not sure either, having concerns that those pull requests who maintaine the lockfile will be flaky.

But that's what we talked about yesterday, and that basically installs the latest transient dependencies in range. As far as i have ready, that's the way renovate is doing this: https://github.com/renovatebot/renovate/discussions/15762

PR #547, opened 2024-11-21
Update deprecated repository methods usages

johnnyomair:
Use the underlying entity manager instead.

### DIFF
--- a/api/src/documents/links/links.resolver.ts
+++ b/api/src/documents/links/links.resolver.ts
@@ -60,14 +60,14 @@ export class LinksResolver {
                 content: input.content.transformToBlockData(),
             });
 
-            this.repository.persist(link);
+            this.repository.getEntityManager().persist(link);
         }
 
         if (attachedPageTreeNodeId) {
             await this.pageTreeService.attachDocument({ id: linkId, type: "Link" }, attachedPageTreeNodeId);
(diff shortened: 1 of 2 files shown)


api/src/documents/links/links.resolver.ts
```diff
@@ -60,14 +60,14 @@ export class LinksResolver {
                 content: input.content.transformToBlockData(),
             });
 
-            this.repository.persist(link);
+            this.repository.getEntityManager().persist(link);
```
johnnyomair, line 63: What do we prefer?

1. Inject entity manager and call `this.entityManager.persist()` (the API Generator does this)
2. Use `repository.getEntityManager()`

johnnyomair: Probably 1️⃣ , but I don't have a strong opinion on this. I've seen both in the core repo.

johnnyomair: Changed: 2e2312f80899b0345ef0581501292ba9d207c73f

johnnyomair:
I've made this a recommended change because it will be required when upgrading to MikroORM 6 / Comet 8.

nsams (approved):

dkarnutsch (approved):

PR #550, opened 2024-11-21
Set engines everywhere

dkarnutsch:
It was odd that we only used it for admin/server anyways...

As per https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines: 
> You can specify the version of node that your stuff works on

Fixes also DO build (see https://docs.digitalocean.com/products/app-platform/reference/buildpacks/nodejs/#current-buildpack-version-and-supported-runtimes): 
> App Platform uses Node.js 20.x by default if you do not specify a Node.js version in the engines section.


johnnyomair (approved):
Lint workflow still needs to be fixed, but looks good

johnnyomair (approved):

nsams (approved):

thomasdax98 (approved):

manuelblum (approved):

PR #565, opened 2024-11-27
Make project starting when site is removed via create-app

fraxachun:
Currently the application does not start anymore when create-app remove-site has been executed due to the deep incorporation of SiteConfigs. This PR injects a lightweight type to comply the needs and is to be meant deleted anyway. This way the application will start with a functioning scope.


johnnyomair (commented):
Wouldn't it be better to remove the site configs from Admin and API when there's no site? Is this even possible?

nsams:
maybe we should get rid of this remove-site instead of doing it "half".

thomasdax98 (approved):

PR #576, opened 2024-11-29
Static rendering

fraxachun:
## Description

https://github.com/vivid-planet/comet/pull/2836 for Starter.

Adds basic styling for NotFound:

image


site/src/app/[domain]/[language]/not-found.tsx
```diff
@@ -0,0 +1,15 @@
+import { getNotFoundContext } from "@src/util/NotFoundContext";
+import Link from "next/link";
+
+export default async function NotFound404(): Promise<JSX.Element> {
+    const scope = getNotFoundContext() || { domain: "main", language: "en" };
+
+    return (
+        <html lang={scope.language}>
+            <body>
+                <p>Page not found (Scope {JSON.stringify(scope)}).</p>
```
Comment (nsams) [inline site/src/app/[domain]/[language]/not-found.tsx:10]: outputting the scope doesn't belong into starter. But localized messages would. (can also be a different PR)

thomasdax98 (approved):

site/src/app/[domain]/[language]/not-found.tsx
```diff
@@ -1,11 +1,28 @@
...
-export default async function NotFound404() {
     return (
-        <html>
+        <html lang={language}>
             <body>
-                <p>Page not found.</p>
-                <Link href="/">Return Home</Link>
+                <IntlProvider locale={language} messages={messages}>
+                    <SiteConfigProvider siteConfig={getSiteConfigForDomain(domain)}>
+                        <Layout params={{ domain, language }}>
```
Comment (johnnyomair) [inline site/src/app/[domain]/[language]/not-found.tsx:19]: Why is this necessary? If the not-found context is defined in the layout and used here, why do we need to render the layout again? Does the not-found context even work?

Comment (johnnyomair) [reply site/src/app/[domain]/[language]/not-found.tsx:19]: I didn't see that, thanks for the explanation! I don't believe the skip was intentional, at least I can't remember why we added it. Maybe we can merge it in a later PR.

johnnyomair (approved):

nsams:
> I had to remove export const dynamic = "force-static";

This is not good, that can easily result in dynamic renders. Isn't there an alternative way to access the cookie? Could a middleware help?

PR #577, opened 2024-11-29
Remove command for removing site

fraxachun:
Reasons to remove this feature:
- It does not make changes in Admin/API which are normally made to transform to a [Data-driven application](https://docs.comet-dxp.com/docs/content-scope/data-driven-application)
- It cannot handly the removal of site-configs properly
- It is prone to not being updated on changes in comet/starter
- There is no automatic process of verifying that the web and/or the deployment works after executing the command


johnnyomair (approved):

thomasdax98:
I'm against this. Though not perfect, this eases the process of removing the site.

PR #581, opened 2024-12-03
Always use mapped type

johnnyomair:
MikroORM will interpret the provided type as column type if no mapped type alias is known:
> The type option here allows several input forms. We are using the text type name here, which is mapped to the TextType - a mapped type representation used internally by the ORM. If you provide a string value there, and it won't match any known type alias, it will be considered as the column type.
> – [MikroORM docs](https://mikro-orm.io/docs/guide/first-entity#scalar-properties)


api/src/db/migrations/Migration20241203124314.ts
```diff
@@ -0,0 +1,14 @@
+import { Migration } from "@mikro-orm/migrations";
+
+export class Migration20241203124314 extends Migration {
+    async up(): Promise<void> {
+        this.addSql('alter table "Footer" alter column "createdAt" type timestamptz(0) using ("createdAt"::timestamptz(0));');
```
johnnyomair, line 5: This changes the precision for seconds from 6 to 0, so no milliseconds will recorded. IMO this is okay, what do you think? Otherwise I can change "timestamp with time zone" to "timestamptz(6)".

api/src/db/migrations/Migration20241203124314.ts
johnnyomair, line 1: Should I merge the three migrations into a single one?

thomasdax98 (dismissed):

api/src/db/migrations/Migration20241203124314.ts
dkarnutsch: I think it is perfectly fine for this use case

api/src/db/migrations/Migration20241203124314.ts
dkarnutsch: Yes - as we want to keep the starter simple and we've decided that it is no problem, if this is breaking

dkarnutsch (dismissed):

nsams (dismissed):

api/src/db/migrations/Migration20241203124314.ts
johnnyomair: Fixed: 75c76db6a57c4f12c305f8092563ac2558fb2bdd

PR #586, opened 2024-12-04
Improve test workflow labels

johnnyomair:
Noticed this while adding the new lint workflow.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ on:
       - next
 
 jobs:
-  lint:
+  test:
     name: Test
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ on:
 
 jobs:
   lint:
```
dkarnutsch, line 18:
```suggestion
  lint:
```

why lint?

johnnyomair: Fixed: e2a2239bf4ca0a10c02614e993947b013a0de5f9

dkarnutsch (approved):

PR #591, opened 2024-12-09
Don't fail GraphQL Codegen if there are no documents

johnnyomair:
This happened in one of our projects: GraphQL Codegen fails with the non-obvious error `A 'const' initializer in an ambient context must be a string or numeric literal or literal enum reference.`. To fix this, we enable the `ignoreNoDocuments` option.

### DIFF
--- a/admin/codegen.ts
+++ b/admin/codegen.ts
@@ -53,6 +53,7 @@ const config: CodegenConfig = {
             ],
         },
     },
+    ignoreNoDocuments: true,
 };
 
 export default config;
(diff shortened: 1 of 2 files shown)


johnnyomair:
Do we even want this? Alternatively, one could remove GraphQL Codegen altogether if there will be never any GraphQL documents.

dkarnutsch (approved):

thomasdax98 (approved):

manuelblum (approved):

PR #596, opened 2024-12-10
Fix warning of props being passed to DOM

fichtnerma:
## Description
Fix a bug that caused props of styled components to be passed to the DOM resulting in an error
- using the shouldForwardProp prevents this

### DIFF
--- a/site/src/common/components/Typography.tsx
+++ b/site/src/common/components/Typography.tsx
@@ -191,11 +191,15 @@ const variantToElementMap: Record<TypographyVariant, "h1" | "h2" | "h3" | "h4" |
     p200: "p",
 };
 
-export const Typography = styled.div.attrs<{
-    as?: unknown;
-    variant?: TypographyVariant;
-    bottomSpacing?: boolean;
-}>((props) => ({ as: props.as ?? variantToElementMap[props.variant ?? "p300"] }))`
+export const Typography = styled.div
+    .withConfig({
+        shouldForwardProp: (prop) => !["variant", "bottomSpacing"].includes(prop),
+    })
+    .attrs<{
+        as?: unknown;
+        variant?: TypographyVariant;
+        bottomSpacing?: boolean;
+    }>((props) => ({ as: props.as ?? variantToElementMap[props.variant ?? "p300"] }))`
     font-family: ${({ theme }) => theme.fontFamily};
     ${({ variant = "p300" }) => typographyVariantStyle[variant]};
     margin-top: 0;


site/src/common/components/Typography.tsx
```diff
@@ -191,11 +191,15 @@ const variantToElementMap: Record<TypographyVariant, "h1" | "h2" | "h3" | "h4" |
...
 };
 
-export const Typography = styled.div.attrs<{
-    as?: unknown;
-    variant?: TypographyVariant;
-    bottomSpacing?: boolean;
-}>((props) => ({ as: props.as ?? variantToElementMap[props.variant ?? "p300"] }))`
+export const Typography = styled.div
+    .withConfig({
+        shouldForwardProp: (prop) => !["variant", "bottomSpacing"].includes(prop),
```
johnnyomair, line 196: Shouldn't `as` be included as well?

nsams:
have you also considered renaming the props to transient props (prefix with `$`)?

fichtnerma:
> have you also considered renaming the props to transient props (prefix with `$`)?

Yes i have also looked into using transient props but decided against it, to not change the interface

site/src/common/components/Typography.tsx
fichtnerma: Including it is not necessary as it is a valid prop for styled components and is consumed https://styled-components.com/docs/api#as-polymorphic-prop

johnnyomair:
> have you also considered renaming the props to transient props (prefix with `$`)?

Changing it to transient props would require devs to add the $ when using the Typography component anywhere, including other components. I'm not sure if we want this.

```tsx
<Typography $variant="..." $bottomSpacing="..." />
```

nsams:
> Changing it to transient props would require devs to add the $ when using the Typography component anywhere

yes, that is really not good to use. It should be consistent across all components.

johnnyomair:
> yes, that is really not good to use. It should be consistent across all components.

What do you mean? Not using transient props at all?

nsams:
yes.

However, our "new" styling library won't have this syntax anyway. (whichever it will be)

johnnyomair (approved):

SebiVPS (approved):

nsams (approved):

manuelblum (approved):

johnnyomair:
@fichtnerma please open a PR for Demo as well, thanks!

fichtnerma:
> Needs a PR in Demo.

Done https://github.com/vivid-planet/comet/pull/3083

johnnyomair (approved):

PR #613, opened 2024-12-15
Knip: Ignore files generated by CRUD Generator

dkarnutsch:
e.g., when using
```
@CrudGenerator({
    targetDirectory: `${__dirname}/../generated/`,
    list: false
})
```
`EntityArgs`, `EntityFilter` or `EntitySort` are still being generated by the CRUD generator, but unused. I think this behaviour from the CRUD generator is perfectly fine, as the CRUD generator does not know if they might be used somewhere else, so we should ignore it.

### DIFF
--- a/admin/knip.json
+++ b/admin/knip.json
@@ -2,6 +2,6 @@
     "$schema": "https://unpkg.com/knip@5/schema.json",
     "entry": ["./src/loader.ts"],
     "project": ["./src/**/*.{ts,tsx}"],
-    "ignore": ["./codegen.ts", "./schema.gql"],
+    "ignore": ["./codegen.ts", "./schema.gql", "./src/**/generated/**/*.{ts,tsx}", "./src/**/*.cometGen.ts"],
     "ignoreDependencies": ["@swc/plugin-emotion"]
 }
--- a/api/knip.json
+++ b/api/knip.json
@@ -2,7 +2,7 @@
     "$schema": "https://unpkg.com/knip@5/schema.json",
     "entry": ["./src/main.ts", "./src/console.ts", "./src/repl.ts", "./src/db/ormconfig.cli.ts", "./src/site-configs.d.ts"],
     "project": ["./src/**/*.{ts,tsx}"],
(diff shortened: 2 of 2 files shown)


nsams:
We don't should generate files that we don't use, I think we should fix that in the generator.

thomasdax98 (dismissed):

dkarnutsch:
> We don't should generate files that we don't use, I think we should fix that in the generator.

I'm not sure about this.

Suppose you want to disable the listQuery because you need custom logic here. But you still might want to use the generated Sort or Filter objects.

Alternatively, we could add flags for all of the objects, but this might be too verbose.

nsams:
> Suppose you want to disable the listQuery because you need custom logic here. But you still might want to use the generated Sort or Filter objects.

Is it Ok to mix scaffolded with generated code? Shouldn't you scaffold the dto files as well in that case?

dkarnutsch:
> > Suppose you want to disable the listQuery because you need custom logic here. But you still might want to use the generated Sort or Filter objects.
> 
> Is it Ok to mix scaffolded with generated code? Shouldn't you scaffold the dto files as well in that case?

I think this is ok, but this might be something to discuss.

manuelblum:
I think we should remove generated code (mainly meaning api-generator, admin-generator generated code). And disable ignore them in linting (eslint / prettier, and knip). 

The generator is in my opinion in the responsibility to generate propper code, and the project, should not care about the generated code style. and if an export is used or not.

As soon as the developer starts to scaffold, the responsibility from the scaffolded code transfers from the generator to the developer, respectively the project, and it should be compliant to the coding standards (eslint, prettier and all other tools).

manuelblum (approved):

nsams (approved):

johnnyomair (commented):
We should probably add the ignore for the Admin Generator as well.

dkarnutsch:
> We should probably add the ignore for the Admin Generator as well.

Done with ae8abd5bd09363d112337e7c7aeb4f3c965a0607.

johnnyomair (approved):

manuelblum (approved):

PR #624, opened 2024-12-19
Convert Next config file to ECMAScript module

SebiVPS:
## Description

https://nextjs.org/docs/pages/api-reference/config/next-config-js#ecmascript-modules

When we want to use ECMA script modules (used by panda-css or next-yak for eg.) we need to update the next-config file from js to mjs.

### DIFF
--- a/site/next.config.mjs
+++ b/site/next.config.mjs
@@ -1,11 +1,12 @@
-/* eslint-disable */
-
 // @ts-check
-const withBundleAnalyzer = require("@next/bundle-analyzer")({
+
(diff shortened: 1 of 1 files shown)


nsams (approved):

nsams:
in next 15 we can use .ts

johnnyomair (approved):

dkarnutsch (approved):

thomasdax98 (approved):

manuelblum (approved):

PR #626, opened 2024-12-19
RFC: next-yak

SebiVPS:
## Description

Replaced styled-components with [next-yak](https://github.com/jantimon/next-yak).

## To Be Discussed:

- I used contants instead of full theme context. `next-yak` would provide [theming](https://yak.js.org/features#theming), but its solved with local cookies. When we use CSS variables for theming, we do not need full theme context support.

Solutions for missing features: 
 - currently `.withConfig` and the `as` prop are not supported. -> Solution to replace `as` usage?
 - no replacement for createGlobalStyle  found (how to solve `DisableBodyScroll` gobal style?

### ToDos:
 - improve `global.css` (how to use yak constants?)

### To test: 
- Usage with Real Server Components
- Performance
- Full test of all components

### Good

- Easy to use (DevXP), same syntax as styled-components.

### Not so good

- missing features (`as` and `createGlobalStyle` - we will find solutions for that)
- Compatibility (we are behind with next and react, so we are forced to use v2 for now)

| next-yak         | Next.js         | react            | swc_core         | 
|------------------|-----------------|------------------|------------------|
| 4.x              | >= 15.0.4       | 19.x             | 5.0.1            |
| 3.x              | 15.x            | 18.x / 19.x      | 3.0.2            |
| 2.x              | 14.x            | 18.x / 19.x      | 0.279.0          |


site/src/constants.yak.ts
```diff
@@ -0,0 +1,49 @@
+export const colors = {
```
SebiVPS, line 1: To even make the migration from the old solution even simpler, we could export here one "theme" object with the same structure like before with all constants, instead of multiple separated objects. Noticed that a bit too late.

johnnyomair:
We've reviewed this today in the Styling JFX and decided that we don't want to invest more effort into evaluating next-yak. The main reason for this decision was the novelty and little adoption of next-yak. Furthermore, as we don't plan to migrate existing projects to the new styling solution, having an easy upgrade path (which is the main argument of next-yak) isn't _that_ important.

PR #639, opened 2024-12-20
Add `ErrorHandler` to site

thomasdax98:
## Description

see https://github.com/vivid-planet/comet/pull/2281

By switching to SSR instead of SSG, an error thrown in a block causes the entire page to crash (white screen). To avoid this, errors in the block are caught and handled in the ErrorHandler. 

In development, the error is still thrown. In production, blocks with errors are simply hidden. A tool like Sentry should be configured where the error can be logged.

### DIFF
--- a/site/src/util/ErrorHandler.tsx
+++ b/site/src/util/ErrorHandler.tsx
@@ -1,0 +1,17 @@
+"use client";
+
+import { ErrorHandlerProvider } from "@comet/cms-site";
+import { PropsWithChildren } from "react";
+
+export function ErrorHandler({ children }: PropsWithChildren) {
+    function onError(error: Error, errorInfo: React.ErrorInfo) {
+        if (process.env.NODE_ENV === "development") {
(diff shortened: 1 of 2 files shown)


nsams (approved):

site/src/util/ErrorHandler.tsx
```diff
@@ -0,0 +1,17 @@
+"use client";
+
+import { ErrorHandlerProvider } from "@comet/cms-site";
+import { PropsWithChildren } from "react";
+
+export function ErrorHandler({ children }: PropsWithChildren) {
+    function onError(error: Error, errorInfo: React.ErrorInfo) {
+        if (process.env.NODE_ENV === "development") {
+            console.error("Error caught by error handler", error, errorInfo.componentStack);
```
johnnyomair, line 9: Do we need this console log if we re-throw the error?

johnnyomair (approved):

dkarnutsch (approved):

manuelblum (approved):

PR #643, opened 2025-01-09
RFC: StyleX

MFSepplive:
We want to replace styled-components with another styling solution to support Server Components. This PR demonstrates how some components, which have dynamic styles, could be implemented using [styleX](https://stylexjs.com/).

This PR is just to start a discussion 😉


johnnyomair:
We've reviewed this today in the Styling JFX and decided that we don't want to invest more effort into evaluating StyleX. The main reason for this decision was the unfamiliar syntax and concepts introduced by StyleX. Furthermore, as @MFSepplive pointed out, StyleX seems to be designed with component libraries in mind, which our projects aren't.

PR #655, opened 2025-01-10
Pin TypeScript version to `5.1.6` (for now)

johnnyomair:
## Description

Running ESLint or any command that uses ESLint (e.g., the API Generator) currently produces the following massive warning:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.2.0

YOUR TYPESCRIPT VERSION: 5.2.2

Please only submit bug reports when using the officially supported version.

=============
```

To fix this, we pin the TypeScript version to the last version supported by typescript-eslint. We then need to update our dependency on typescript-eslint in the library and correctly set a peer dependency on TypeScript in `@comet/eslint-config`.

## Further information

A PR in Demo isn't necessary because Demo still uses TypeScript 4.

### DIFF
--- a/admin/package-lock.json
+++ b/admin/package-lock.json
@@ -67,7 +67,7 @@
                 "prettier": "^2.8.8",
                 "rimraf": "^6.0.1",
                 "ts-node": "^10.9.2",
-                "typescript": "^5.7.2",
+                "typescript": "5.1.6",
                 "vite": "^5.4.11",
                 "vite-plugin-html": "^3.2.2"
             },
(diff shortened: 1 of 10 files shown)


nsams:
maybe starter+library should all use the same ts version to avoid such issues

nsams (approved):

dkarnutsch (approved):

thomasdax98 (approved):

PR #670, opened 2025-01-15
Add default values for preliminary domain and prelogin

fraxachun:
- Preliminary Domain is only necessary for prod.
- Prelogin is only necessary for nonprod environments (exception prod in conjunction with preliminary, but this can't be handled in site-configs).


johnnyomair (commented):
> exception prod in conjunction with preliminary, but this can't be handled in site-configs

This is our normal use case for all new projects – how would we handle this? Also, what would happen if devs would forget to adapt this in site configs? Would the site be public, or would this be prevented in another location?

johnnyomair:
Thanks for the detailed explanation!

> Since the latter is worse and we normally use a preliminary domain I decided for this PR.

Unfortunately, this isn't the case for all our projects. How would we proceed for them?

thomasdax98 (approved):

manuelblum (approved):

dkarnutsch:
> but this can't be handled in site-configs

That's a pity, because I think the current solution is a bit misleading. On first sight it seems like a bug. Maybe we should add a comment here?

dkarnutsch (approved):

johnnyomair:
> The best solution would be if the name reflects the logic more, e.g. domains.prelogin instead domains.preliminary. But that would be breaking. So I added a comment: https://github.com/vivid-planet/comet-starter/commit/625efe2255c1ce914875e7223e00d1f86fa6e8a8

Should we rename this in core in next?

johnnyomair (approved):

PR #671, opened 2025-01-15
Refactor single middleware function into multiple functions that get chained together

nsams:
This avoids having one large file doing multiple jobs.

Based on https://github.com/vivid-planet/comet/pull/2728 (and others)


nsams:
> Or do you prefer doing such improvements in demo first?

yes, better start a new PR for future improvements. Can you create one in demo for that?

johnnyomair (dismissed):

dkarnutsch (dismissed):

site/src/middleware/cspHeaders.ts
```diff
@@ -0,0 +1,34 @@
+import { NextRequest } from "next/server";
+
+import { CustomMiddleware } from "./chain";
+
+export function withCspHeadersMiddleware(middleware: CustomMiddleware) {
```
manuelblum, line 5:
```suggestion
export function withContentSecurityPolicyHeadersMiddleware(middleware: CustomMiddleware) {
```

site/src/middleware/damRewrite.ts
```diff
@@ -0,0 +1,12 @@
+import { NextRequest, NextResponse } from "next/server";
+
+import { CustomMiddleware } from "./chain";
+
+export function withDamRewriteMiddleware(middleware: CustomMiddleware) {
```
manuelblum, line 5: Due to this is a very essential part of the application - every request goes through those middelwares, I would prefer to have some more information / documentation on what the middlewares are doing. Even though this is  function is indeed very simple (the others in my opinion are more advanced) i would prefer to have something like this for the middlewares:

```suggestion
/**
* This function rewrites requests starting with `/dam` to the internal resource location
*/
export function withDamRewriteMiddleware(middleware: CustomMiddleware) {
```

manuelblum (approved):

thomasdax98 (dismissed):

site/src/middleware/damRewrite.ts
nsams: I try to avoid comments that give no additional information. Instead I give my best with self explaining function names. 

DomainRewrite really needs explanation, I'll add a comment. DamRewrite imho doesn't.

johnnyomair (approved):

manuelblum (dismissed):

johnnyomair:
@nsams please fix conflicts.

johnnyomair (approved):

PR #690, opened 2025-01-24
Pin version of Comet packages

thomasdax98:
## Background

During COMET development, we assume that the installed version of comet packages is identical across all microservices. Thus having a different version of, e.g., cms-api and cms-admin could cause unforeseen problems.

## Problem

Defining the version with a caret like this `^7.8.0` means all newer minor and patch versions also match. So, it's possible to have version 7.8.0 of `@comet/cms-admin` but 7.12.0 of `@comet/cms-api`.

Installing different version can happen accidentally (just recently happened in a client project). A mistake like that is hard to spot because you have to look into the package-lock. 

## Solution

I propose pinning the version of comet packages to prevent such mistakes.


admin/package.json
```diff
@@ -27,11 +27,11 @@
     },
     "dependencies": {
         "@apollo/client": "^3.11.9",
-        "@comet/admin": "^7.12.0",
-        "@comet/admin-icons": "^7.12.0",
-        "@comet/admin-theme": "^7.12.0",
-        "@comet/blocks-admin": "^7.12.0",
-        "@comet/cms-admin": "^7.12.0",
+        "@comet/admin": "7.12.0",
```
manuelblum, line 30: I think it's a good change 👍 and makes things better.

Do we get problems when `api <-> admin`, respectively `site <-> api` have different comet versions?

manuelblum (dismissed):

nsams:
how do you update to the lastest comet patch version?

admin/package.json
thomasdax98: > Do we get problems when api <-> admin, respectively site <-> api have different Comet versions?

Yes that's possible if there were changes to the GraphQL API. 

> It's probably hard to enforce this behaviour.

I don't think there's a way to do this with npm. We recommend updating together and assume that all packages have the same version. If we wouldn't do this, it would basically prevent us from making any changes to the API in minor version.

thomasdax98:
> how do you update to the lastest comet patch version?

Like any other update by increasing the version number in the package.json. I don't really see a difference to minor version updates. 

Also once https://github.com/vivid-planet/comet-upgrade/pull/51 and https://github.com/vivid-planet/comet-upgrade/pull/52 are merged, I'd like to start recommending to do all updates with `@comet/upgrade`

nsams:
> Like any other update by increasing the version number in the package.json.

Doing that manually is no real option imho. If `@comet/upgrade` supports that properly, then Ok.

admin/package.json
manuelblum: I am with you, just wanted to mention it that this could be problematic

johnnyomair (approved):

dkarnutsch (approved):

johnnyomair (changes requested):
Some dependency updates have been reverted (probably due to a merge error). Please fix.

johnnyomair (approved):

PR #705, opened 2025-01-30
Present 404 when domain lookup not successful

fraxachun:
Implements https://github.com/vivid-planet/comet/pull/3300

Shows an empty page as this 404 is just a fallback for missing configurations.

### DIFF
--- a/site/src/documents/pages/Page.tsx
+++ b/site/src/documents/pages/Page.tsx
@@ -131,7 +131,7 @@ export async function Page({ pageTreeNodeId, scope }: { pageTreeNodeId: string; 
     return (
         <>
             {document.seo.structuredData && document.seo.structuredData.length > 0 && (
-                <script type="application/ld+json">{document.seo.structuredData}</script>
+                <script type="application/ld+json" dangerouslySetInnerHTML={{ __html: document.seo.structuredData }} />
             )}
             <main>
                 <StageBlock data={document.stage} />


site/src/app/not-found.tsx
johnnyomair, line 1: Do we need this? We could also use Next' built-in 404 for this special case

manuelblum (approved):

nsams (approved):

site/src/app/not-found.tsx
nsams: interesting technique

but now you need to hardcode a default domain and language in the middleware.... i'm not sure if that is better

johnnyomair: Alternative: Present a JSON response with 404, like we do for status.

johnnyomair (dismissed):

thomasdax98 (approved):

manuelblum (dismissed):

dkarnutsch (dismissed):

nsams:
I haven't approved yet, becuase we didn't decide on  `new NextResponse('<html><body>404 Can not resolve domain</body></html>', { status: 404 });` vs  `new NextResponse(null, { status: 404 });`

site/src/middleware/redirectToMainHost.ts
```diff
@@ -23,7 +23,14 @@ export function withRedirectToMainHostMiddleware(middleware: CustomMiddleware) {
...
 
-            throw new Error(`Cannot get siteConfig for host ${host}`);
+            const response = new NextResponse(
+                '<!DOCTYPE html><html lang="en"><head><meta charset=utf-8><title>404</title></head><body>Cannot resolve domain</body></html>',
+                {
+                    status: 404,
+                },
+            );
+            response.headers.set("Content-Type", "text/html; charset=utf-8");
+            return response;
```
johnnyomair, line 33:
```suggestion
            return NextResponse.json({ error: "Cannot resolve domain" }, { status: 404 })
```

I'd suggest returning JSON instead like we do for the site preview: https://github.com/vivid-planet/comet-starter/blob/main/site/src/middleware/sitePreview.ts#L14

johnnyomair: (It's easier and consistent).

nsams (dismissed):

nsams (approved):

johnnyomair (approved):

PR #733, opened 2025-02-17
Use base64 for encoding site-configs

fraxachun:
See https://github.com/vivid-planet/comet/pull/3053


nsams (approved):

johnnyomair (commented):
One downside of this approach is that devs can't easily review the local env variable during development. Are we okay with this? Decoding base64 is simple, but may still be an inconvenience

johnnyomair (approved):

dkarnutsch (approved):

thomasdax98 (approved):

manuelblum (approved):

PR #794, opened 2025-03-24
Draft: Update to COMET v8.0.0-beta.2

thomasdax98:


admin/package.json
```diff
@@ -28,15 +28,15 @@
...
+        "@comet/cms-admin": "8.0.0-beta.2",
         "@emotion/react": "^11.14.0",
         "@emotion/styled": "^11.14.0",
         "@fontsource-variable/roboto-flex": "^5.2.5",
-        "@mui/material": "^6.4.8",
-        "@mui/x-data-grid": "^7.28.0",
-        "@mui/x-data-grid-pro": "^7.28.0",
+        "@mui/material": "^6.0.0",
+        "@mui/x-data-grid": "^7.0.0",
+        "@mui/x-data-grid-pro": "^7.0.0",
```
johnnyomair, line 39: It isn't optimal that the upgrade scripts revert correct changes. Maybe we can change the dependency updates to check if the existing version already matches the target version.

johnnyomair (dismissed):

admin/package.json
thomasdax98: https://github.com/vivid-planet/comet-upgrade/pull/85

By now I'm (again) more in favor of tracking executed upgrade scripts and not executing them multiple times (instead of making them idempotent) because

1. It's very easy to overlook such details (even when scripts are executed twice during development)
2. We have a lot of upgrade scripts for v8 and executing all of them is already very slow

johnnyomair (approved):

admin/package.json
johnnyomair: > By now I'm (again) more in favor of tracking executed upgrade scripts and not executing them multiple times (instead of making them idempotent) because

Could we discuss this in this week's meeting?

thomasdax98: Yes

PR #803, opened 2025-03-24
Update create-app to COMET v8

thomasdax98:
## Description

- Update eslint, prettier and comet in create-app
- Require Node 22 (instead of Node 20)

I tested by creating a new project. It starts and the lint passes


create-app/package.json
```diff
@@ -32,22 +32,23 @@
...
-        "@tsconfig/node20": "^20.1.4",
+        "@tsconfig/node22": "^22.0.1",
         "@types/node": "^22.13.10",
         "cspell": "^8.17.5",
         "npm-run-all2": "^7.0.2",
-        "prettier": "^2.8.8",
-        "typescript": "5.8.2"
+        "prettier": "^3.5.3",
+        "typescript": "5.8.2",
+        "eslint-plugin-package-json": "0.26.3"
```
thomasdax98, line 51: This is temporary and only necessary because they made a breaking change in 0.26.4: https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/compare/v0.26.3...v0.26.4#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L25-L31

thomasdax98: Will be fixed with https://github.com/vivid-planet/comet/pull/3689

create-app/bin/index.js
johnnyomair, line 1: This file probably shouldn't be empty.

thomasdax98: The file isn't actually empty. I only (accidentally) changed the permissions from 755 to 644 (I think because I deleted and recreated the file).

johnnyomair (approved):

PR #817, opened 2025-03-26
Fix broken site page for unknown language in URL

manuelblum:
Opening an URL with an unknown language leads to an runtime error in the page.

**before:**
Screenshot 2025-03-26 at 11 22 58

**after:**
Screenshot 2025-03-26 at 11 22 32

## Further information

Ticket: https://vivid-planet.atlassian.net/browse/COM-1868


johnnyomair (commented):
Do we need to fix this in Demo as well?

manuelblum:
> Do we need to fix this in Demo as well?

not needed, its already implemented in demo: https://github.com/vivid-planet/comet/blob/next/demo/site/src/app/%5Bvisibility%5D/%5Bdomain%5D/%5Blanguage%5D/%5B%5B...path%5D%5D/layout.tsx#L16-L19

johnnyomair (approved):

PR #824, opened 2025-03-31
Adjust DigitalOcean App Platform oauth2-proxy configuration to meet our best practices

dkarnutsch:
## Description

- adds skip auth route to make /status for monitoring available
- reduce cookie refresh
- reorder values and remove reconfiguration of default values


johnnyomair (approved):
> reorder values and remove reconfiguration of default values

This could have been separate commits, then reviewing the actual changes would have been easier.

PR #825, opened 2025-03-31
Fixes for Docker deployment

dkarnutsch:
## Description

- add cert directory to gitignore to keep working tree clean
- add migration command to startup
- add missing imgproxy secrets file
- add hint about installing root deps
- harmonize oauth2-proxy config with best practices

### DIFF
--- a/.docker-compose/docker-compose.tpl.yml
+++ b/.docker-compose/docker-compose.tpl.yml
@@ -30,6 +30,7 @@ services:
       IMGPROXY_USE_S3: "true"
       IMGPROXY_S3_REGION: fra1
       IMGPROXY_S3_ENDPOINT: https://comet-starter.fra1.digitaloceanspaces.com
+    env_file: "imgproxy.secrets.env"
 
   oauthproxy:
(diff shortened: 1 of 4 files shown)


.docker-compose/docker-compose.tpl.yml
```diff
@@ -30,34 +30,42 @@ services:
       IMGPROXY_USE_S3: "true"
       IMGPROXY_S3_REGION: fra1
       IMGPROXY_S3_ENDPOINT: https://comet-starter.fra1.digitaloceanspaces.com
+    env_file: "imgproxy.secrets.env"
```
johnnyomair, line 33: Where does this file come from?

dkarnutsch: Similar to the other secrets files: you have to create them on your server and they are gitignored.

johnnyomair (approved):

.docker-compose/docker-compose.tpl.yml
dkarnutsch: This might be added to the docs as well...

PR #826, opened 2025-03-31
Use static user as long static user service is used

fraxachun:
https://vivid-planet.atlassian.net/browse/COM-1870

The starter implementation by default uses the `StaticUsersUserService` for the `UserPermissionsModule`. In contrary, the deployment templates (and hence our demo deployments) we provide use an Authproxy, for this reason we use `createJwtAuthService` in the `AuthService`. The problem is that the `CometAuthGuard` retrieves the authed user id from `createJwtAuthService` and then tries to load the user from the `StaticUsersUserService` which results in a mismatch.

This PR works around this problem by providing `convertJwtToUser` which blindly loads a static user as long the appropriate user service is not used.

Further changes of this PR:
- Rename `UserService` to `StaticUsersUserService`
- Change names of static users to show clearly that static users are used
- Add todos for devs what to change when implementing the correct user service

### DIFF
--- a/api/src/app.module.ts
+++ b/api/src/app.module.ts
@@ -27,7 +27,7 @@ import { Request } from "express";
 
 import { AccessControlService } from "./auth/access-control.service";
 import { AuthModule, SYSTEM_USER_NAME } from "./auth/auth.module";
-import { UserService } from "./auth/user.service";
+import { StaticUsersUserService } from "./auth/static-users.user.service";
 import { Config } from "./config/config";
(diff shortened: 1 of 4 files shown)


dkarnutsch:
The rename should also happen for demo. Maybe after https://github.com/vivid-planet/comet/pull/3742/.

dkarnutsch (approved):

johnnyomair (approved):

PR #827, opened 2025-04-01
Rename API_PASSWORD env for site

dkarnutsch:
See https://github.com/vivid-planet/comet/pull/3748. This is also part of the reason, why currently the deployment on DigtialOcean is not working.


johnnyomair (commented):
We should decide if we want to throw an error first.

dkarnutsch:
> We should decide if we want to throw an error first.

https://github.com/vivid-planet/comet/pull/3748 is now merged.

johnnyomair (approved):

PR #842, opened 2025-04-29
Site: Fix breakpoints

SebiVPS:
## Description

The breakpoints in the site theme were incorrect: xs was defined as 600 px, while it actually is 0-599 px in the design library. I have adjusted the breakpoints in the theme and checked the usages against the design library, verifying that all usages are correct.

https://github.com/vivid-planet/comet/pull/3755

### DIFF
--- a/site/src/common/blocks/MediaGalleryBlock.tsx
+++ b/site/src/common/blocks/MediaGalleryBlock.tsx
@@ -53,15 +53,15 @@ const PageLayoutContent = styled.div`
     grid-column: 2 / -2;
     position: relative;
 
-    ${({ theme }) => theme.breakpoints.xs.mediaQuery} {
+    ${({ theme }) => theme.breakpoints.sm.mediaQuery} {
         grid-column: 5 / -5;
     }
 
-    ${({ theme }) => theme.breakpoints.md.mediaQuery} {
+    ${({ theme }) => theme.breakpoints.lg.mediaQuery} {
         grid-column: 6 / -6;
     }
 
-    ${({ theme }) => theme.breakpoints.lg.mediaQuery} {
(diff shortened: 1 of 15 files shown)


site/src/common/blocks/MediaGalleryBlock.tsx
```diff
@@ -73,7 +73,7 @@ const MediaCaption = styled(Typography)`
     /* min-height to show arrows when no caption */
     min-height: 20px;
 
-    ${({ theme }) => theme.breakpoints.md.mediaQuery} {
+    ${({ theme }) => theme.breakpoints.lg.mediaQuery} {
```
johnnyomair, line 76: This wasn't changed in Demo, is this correct?

site/src/common/components/Typography.tsx
johnnyomair, line 1: This wasn't changed in Demo, is this correct?

site/src/layout/header/Header.tsx
johnnyomair, line 1: This wasn't changed in Demo, is this correct?

site/src/layout/header/MobileMenu.tsx
johnnyomair, line 1: This wasn't changed in Demo, is this correct?

johnnyomair (commented):
The change between the PRs in Demo and here aren't the same. Could you please compare your additional changes with the design library to see if they're correct? Thanks!

site/src/layout/header/MobileMenu.tsx
SebiVPS: Well, there is no such file in Demo ...

site/src/layout/header/Header.tsx
SebiVPS: yes

johnnyomair (approved):

PR #848, opened 2025-05-15
Add 1200x630 to allowed aspect ratios

thomasdax98:
As done here in Demo: https://github.com/vivid-planet/comet/pull/2062/files#diff-bb62086b1bdcdc241c952ff819e5915c6abd367710e871998eafc5768312a3a2R13

### Why?

We use 1200 x 630 for og images (see https://github.com/vivid-planet/comet-starter/blob/a13ac48cd35fada990c361f0b3c0b90a181d2cfa/site/src/documents/pages/Page.tsx#L78). Currently this is broken in starter because the aspect ratio is not allowed


johnnyomair (approved):
I can't explain how we didn't notice this for such a long time 😅

PR #896, opened 2025-07-02
Fix Knip for Admin server

johnnyomair:
## Description

Error introduced with https://github.com/vivid-planet/comet-starter/pull/892. Adapt Knip config to detect used libraries correctly.

### DIFF
--- a/admin/knip.json
+++ b/admin/knip.json
@@ -8,7 +8,6 @@
             "project": ["./src/**/*.{ts,tsx}"]
         },
         "server": {
-            "ignoreDependencies": ["http-proxy-middleware"],
             "project": ["**/*.js"]
         }
     }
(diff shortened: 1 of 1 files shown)


admin/knip.json
```diff
@@ -3,12 +3,12 @@
     "workspaces": {
```
manuelblum, line 3: tested this locally with this configuration:

```
{
    "$schema": "https://unpkg.com/knip@5/schema.json",
    "workspaces": {
        ".": {
            "entry": ["./src/loader.ts"],
            "ignore": ["./codegen.ts", "./schema.gql", "./src/**/generated/**/*.{ts,tsx}", "./src/**/*.cometGen.ts"],
            "ignoreDependencies": ["@swc/plugin-emotion"],
            "project": ["./src/**/*.{ts,tsx}"]
        },
        "server": {
            "project": ["**/*.js"]
        }
    }
}
```
and it seems to me that it works correctly then.

PR #1329, opened 2026-03-06
Add `lint:fix` Commands and Update Agent Guidance

jamesricky:
## Description

Previously there was no consistent way to auto-fix linting and formatting issues — only lint check commands existed. 

This adds `lint:fix` scripts to api, admin, and site, as well as a root-level script that runs all three in parallel. 
Each script wires together ESLint `--fix,` Prettier `--write`, and (for site) stylelint `--fix`. 
`AGENTS.md` is updated to document these commands and to instruct agents to run `lint:fix` after making code changes, keeping the codebase consistently formatted without manual intervention.


johnnyomair (dismissed):
We should add this in core as well.

thomasdax98 (dismissed):

thomasdax98 (dismissed):

johnnyomair (commented):
Should we add a lint:fix script for create-app as well?

johnnyomair (changes requested):

johnnyomair (approved):

PR #1494, opened 2026-06-08
Update COMET to v9.0.0-beta.5

thomasdax98:
Update from v9.0.0-beta.0 to v9.0.0-beta.5


admin/src/documents/pages/EditPage.tsx
```diff
@@ -121,7 +116,9 @@ export const EditPage = ({ id }: Props) => {
             {hasChanges && (
                 <RouterPrompt
                     message={(location) => {
-                        if (location.pathname.startsWith(match.url)) return true; //we navigated within our self
+                        if (location.pathname.startsWith(match.url)) {
+                            return true;
+                        } //we navigated within our self
```
johnnyomair:
```suggestion
                        if (location.pathname.startsWith(match.url)) {
                            //we navigated within our self
                            return true;
                        }
```

johnnyomair (approved):

PR #1495, opened 2026-06-08
Support running multiple dev instances on offset ports

Fabian-Fynn:
Lets several checkouts run npm run dev in parallel without host port collisions — both Docker containers and Node services pick up offset ports.

  Changes
  - assign-ports.sh — reads every *_PORT from .env, finds a free +100..+900 offset, writes base+offset to .env.local (leaves .env untouched)
  - package.json — wrap dev-pm start in dotenv -c secrets so the cascaded env (incl. .env.local) is exported; docker compose prefers those real env vars over its .env file, so containers bind the offset host ports
  - dev-oidc-provider.config.mts — honor IDP_PORT instead of the hardcoded 8080
  - AGENTS.md — document the workflow

  Usage
  ./assign-ports.sh   # picks a free offset → .env.local
  npm run dev


nsams:
- I would prefer typescript instead of bash for non-trivial code.
- IDP_PORT fix should be in it's own PR
- usually we do demo first

johnnyomair:
> IDP_PORT fix should be in it's own PR

Done for Demo here: https://github.com/vivid-planet/comet/pull/5852. Will use this to try @manuelblum's needs-starter-pr-routine.

johnnyomair (commented):
Please update this PR to your updated version once that's merged. Thanks!

Fabian-Fynn:
The script is up to date.
I removed the IDP_PORT fix since it's fixed in main

package.json
```diff
@@ -3,7 +3,7 @@
     "private": true,
     "scripts": {
         "create-site-configs-env": "dotenv -e .env.secrets -e .env.local -e .env -- npx @comet/cli inject-site-configs -f site-configs/site-configs.ts -i .env.site-configs.tpl -o .env.site-configs --base64",
-        "dev": "npm run create-site-configs-env && dev-pm start",
+        "dev": "npm run create-site-configs-env && dotenv -c secrets -- dev-pm start",
```
johnnyomair, line 7: This is necessary for Docker to pick up the overridden ports. However, this sets the variables for all scripts in dev-pm config. I'd prefer only setting it for Docker. Maybe we need a `dev:docker` script? Also, `-c secrets` isn't necessary, simply `dotenv` should be enough.

assign-ports.sh
johnnyomair, line 1: I agree with @nsams that a TS script would be easier to comprehend, but I'm fine with a bash script as well.

johnnyomair:
@nsams once suggested the following alternative:

We could use a single `PORT_OFFSET` variable that's used all the time:

```sh
# .env
PORT_OFFSET=0

API_PORT=4$PORT_OFFSET00
ADMIN_PORT=8$PORT_OFFSET00
```

Your script would then only write the determined `PORT_OFFSET` variable to .env.local.

I like this idea. What do you think about it?

VPS-Fabi:
@copilot resolve the merge conflicts in this pull request

VPS-Fabi:
> @nsams once suggested the following alternative:
> 
> We could use a single `PORT_OFFSET` variable that's used all the time:
> 
> ```shell
> # .env
> PORT_OFFSET=0
> 
> API_PORT=4$PORT_OFFSET00
> ADMIN_PORT=8$PORT_OFFSET00
> ```
> 
> Your script would then only write the determined `PORT_OFFSET` variable to .env.local.
> 
> I like this idea. What do you think about it?

Good idea

VPS-Fabi:
@nsams, @johnnyomair 
its now a Typescript script. I also changed it to the single `PORT_OFFSET` which required the port changes.

nsams:
When it is a single variable in .env.local is the Script still needed?

VPS-Fabi:
> When it is a single variable in .env.local is the Script still needed?

It's not needed necessarily. You can set the ports manually. In my opinion it's still useful for finding a free port offset automatically.

assign-ports.mts
```diff
@@ -0,0 +1,115 @@
...
+        const free = await new Promise<boolean>((resolve) => {
+            const server = createServer();
+            server.once("error", (error: NodeJS.ErrnoException) => resolve(error.code !== "EADDRINUSE"));
+            server.once("listening", () => server.close(() => resolve(true)));
+            server.listen(port, host);
+        });
+        if (!free) return false;
+    }
+    return true;
+}
```
VPS-Obi, line 52: We could use the [get-port](https://www.npmjs.com/package/get-port) package for this, but the custom implementation is fine as well.

package.json
VPS-Obi: Please don't call dotenv here. Instead, change the docker script in the dev-pm config to `set -a; . .env; . .env.local; set +a; docker compose up` (we did this in Demo and projects).

VPS-Obi:
As an alternative to the script, we could instruct agents to find the offset themselves, for instance with https://github.com/sindresorhus/get-port-cli. However, I'm don't not sure how reliable this will work.

VPS-Fabi:
> As an alternative to the script, we could instruct agents to find the offset themselves, for instance with https://github.com/sindresorhus/get-port-cli. However, I'm don't not sure how reliable this will work.

I don't see a benefit in using an agent when a script can do it.

PR #1520, opened 2026-06-11
Admin Server: Pin @types/node dependency

manuelblum:
Source PR: https://github.com/vivid-planet/comet/pull/5197

## Description

Since there is no direct dependency on `@types/node` in the server, lock file maintenance could upgrade it to the latest version. Since we're using Node 24 in production, it's a good idea to ensure we have the correct types, even if we're not using them (yet).

## Summary

- `admin/server/package.json`: Added `@types/node` as a `devDependency` pinned to `^24.0.0` (matching the `"node": "24"` engine constraint).
- `admin/server/package-lock.json`: Updated to include `@types/node@24.10.13` and its transitive dependency `undici-types@7.16.0`.

## Skipped

- `pnpm-lock.yaml` changes — Comet uses pnpm; the starter uses npm. The lockfile has been updated separately for npm.

---
_Synced from [vivid-planet/comet#5197](https://github.com/vivid-planet/comet/pull/5197)_

---
_Generated by [Claude Code](https://claude.ai/code/session_01XwTgihUBoKMcyNx95UDzBB)_

### DIFF
--- a/admin/server/package.json
+++ b/admin/server/package.json
@@ -9,5 +9,8 @@
     },
     "engines": {
         "node": "24"
+    },
+    "devDependencies": {
+        "@types/node": "^24.0.0"
     }
 }
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,7 +38,7 @@ jobs:
           registry-url: "https://registry.npmjs.org"
 
       - name: Install dependencies
-        run: npm ci && npm --prefix api ci && npm --prefix admin ci && npm --prefix site ci && npm --prefix create-app ci
+        run: npm ci && npm --prefix api ci && npm --prefix admin ci && npm --prefix admin/server ci && npm --prefix site ci && npm --prefix create-app ci
         env:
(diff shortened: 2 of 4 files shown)


admin/server/package.json
```diff
@@ -9,5 +9,8 @@
     },
     "engines": {
         "node": "24"
+    },
+    "devDependencies": {
+        "@types/node": "^24.0.0"
```
manuelblum, line 14: this should change package-lock - normally npm install pipeline should complain here.

manuelblum: Good catch. The `package-lock.json` has been updated in this PR. The root cause was that `npm ci` does not fail when a new devDependency is missing from the lockfile — it only errors on version conflicts. The CI pipeline also didn't install `admin/server` at all, so the lockfile there was never checked.

The fix — using `npm install && git diff --exit-code` to detect stale lockfiles for all services — is handled in a separate PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

johnnyomair (approved):

PR #1523, opened 2026-06-11
ci: Verify lockfiles are up to date after install

manuelblum:
## Problem

A `package.json` change that adds a dependency without updating the lockfile could pass CI undetected:

- The `admin/server` service was **completely absent** from the CI install steps, so its lockfile was never validated at all.
- Even after adding it, `npm ci` alone does **not** catch the issue: `admin`'s `postinstall` hook runs `cd server && npm install`, and `npm install` (unlike `npm ci`) **silently rewrites** `admin/server/package-lock.json` to add the missing dependency. By the time `npm --prefix admin/server ci` runs, the lockfile is already back in sync, so it passes.

## Solution

Add `admin/server` to the **Install dependencies** step, and add a dedicated **Verify lockfiles are up to date** step right after install in both `lint.yml` and `test.yml`:

```yaml
- name: Verify lockfiles are up to date
  run: git diff --exit-code -- ':(glob)**/package-lock.json'
```

Because the install step (via the postinstall `npm install`) modifies any out-of-sync lockfile in the working tree, `git diff --exit-code` reliably catches it and fails the run. This covers all lockfiles: root, `api`, `admin`, `admin/server`, `site`, and `create-app`.

## Verification

Confirmed end-to-end by temporarily adding a dependency to `admin/server/package.json` without updating the lockfile:

- Before the verify step: the **Test** job stayed green (the postinstall masked the out-of-sync lockfile from `npm ci`).
- After the verify step: the **Test** job failed at *Verify lockfiles are up to date*, with the diff showing `is-odd` written into `admin/server/package-lock.json`.

The temporary dependency was then reverted.

## References

- Jira: [COM-3009](https://vivid-planet.atlassian.net/browse/COM-3009) (epic: [COM-260](https://vivid-planet.atlassian.net/browse/COM-260))

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tga83toTr33QRZi4vYbRAP)_

[COM-3009]: https://vivid-planet.atlassian.net/browse/COM-3009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,41 @@
+name: "Setup"
+description: "Set up Node.js, install dependencies, and verify lockfiles are up to date"
+
+inputs:
+  setup-ci:
+    description: "Run the project's CI setup scripts"
+    default: "true"
+  build:
+    description: "Build all packages"
+    default: "false"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Use Node.js 24.x
+      uses: actions/setup-node@v6
+      with:
+        node-version: 24
+        cache: "npm"
+        cache-dependency-path: "**/package-lock.json"
+        registry-url: "https://registry.npmjs.org"
+
+    - name: Install dependencies
+      shell: bash
+      run: npm ci && npm --prefix api ci && npm --prefix admin ci && npm --prefix admin/server ci && npm --prefix site ci && npm --prefix create-app ci
+      env:
+        NPM_CONFIG_STRICT_PEER_DEPS: true
+
+    - name: Verify lockfiles are up to date
+      shell: bash
+      run: git diff --exit-code -- ':(glob)**/package-lock.json'
+
+    - name: Setup project for CI run
+      if: ${{ inputs.setup-ci == 'true' }}
+      shell: bash
+      run: npm run setup:ci && npm --prefix admin run setup:ci && npm --prefix site run setup:ci
+
+    - name: Build
+      if: ${{ inputs.build == 'true' }}
+      shell: bash
+      run: npm run build && npm --prefix create-app run build
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,14 +1,6 @@
 name: "Setup"
 description: "Set up Node.js, install dependencies, and verify lockfiles are up to date"
 
-inputs:
-  setup-ci:
-    description: "Run the project's CI setup scripts"
-    default: "true"
-  build:
-    description: "Build all packages"
-    default: "false"
-
 runs:
   using: "composite"
   steps:
@@ -31,11 +23,5 @@ runs:
       run: git diff --exit-code -- ':(glob)**/package-lock.json'
 
     - name: Setup project for CI run
-      if: ${{ inputs.setup-ci == 'true' }}
       shell: bash
       run: npm run setup:ci && npm --prefix admin run setup:ci && npm --prefix site run setup:ci
-
-    - name: Build
-      if: ${{ inputs.build == 'true' }}
-      shell: bash
-      run: npm run build && npm --prefix create-app run build
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,8 +31,9 @@ jobs:
 
       - name: Setup
         uses: ./.github/actions/setup
-        with:
-          build: "true"
 
       - name: Lint
         run: npm run lint:ci && npm --prefix api run lint:ci && npm --prefix admin run lint:ci && npm --prefix site run lint:ci && npm --prefix create-app run lint
+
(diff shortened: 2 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -42,6 +42,15 @@ jobs:
         env:
           NPM_CONFIG_STRICT_PEER_DEPS: true
 
+      - name: Verify lockfiles are up to date
+        run: |
+          npm install && git diff --exit-code package-lock.json
```
manuelblum: why do we need to install twice, see some lines in the install dependencies job.

manuelblum:
This was discovered while reviewing #1520, where `@types/node` was added to `admin/server/package.json` without updating the lockfile — and CI passed anyway.

The root cause is that `npm ci` only errors when a package in the lockfile has the wrong version. It does **not** fail when a package is listed in `package.json` but missing from the lockfile entirely. The `admin/server` package was also completely absent from the CI install steps, so its lockfile was never touched at all.

The fix uses `npm install` (which regenerates the lockfile if needed) followed by `git diff --exit-code` to assert no changes occurred. If the lockfile was out of date, the diff will be non-empty and the step fails.

---
_Generated by [Claude Code](https://claude.ai/code)_

nsams:
pnpm > npm

.github/workflows/test.yml
```diff
@@ -38,10 +38,13 @@ jobs:
...
 
       - name: Install dependencies
-        run: npm ci && npm --prefix api ci && npm --prefix admin ci && npm --prefix site ci
+        run: npm ci && npm --prefix api ci && npm --prefix admin ci && npm --prefix admin/server ci && npm --prefix site ci
         env:
           NPM_CONFIG_STRICT_PEER_DEPS: true
 
+      - name: Verify lockfiles are up to date
+        run: git diff --exit-code -- ':(glob)**/package-lock.json'
+
```
johnnyomair: IMO it would be okay if only the lint workflow would check this.

johnnyomair (dismissed):

.github/workflows/test.yml
manuelblum: i think it is related to the install step and not the the lint/test exclusively.

Maybe an overkill, but what if we setup a common action, similar to what we did here.: https://github.com/vivid-planet/comet/pull/5715/changes

manuelblum: Done in 021fd76. Added a reusable composite action at `.github/actions/setup/action.yml` (Node.js setup → install → lockfile verification → optional `setup:ci`/`build`) and both `lint.yml` and `test.yml` now just `uses: ./.github/actions/setup`. So the install + lockfile check lives in one shared place rather than being duplicated per workflow.

Kept it npm-based for this PR. Left `copilot-setup-steps.yml` as-is since it uses an intentional parallel install and an extra agent-skills step — happy to fold it in too if you'd like.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/lint.yml
```diff
@@ -29,24 +29,10 @@ jobs:
...
-          registry-url: "https://registry.npmjs.org"
-
-      - name: Install dependencies
-        run: npm ci && npm --prefix api ci && npm --prefix admin ci && npm --prefix site ci && npm --prefix create-app ci
-        env:
-          NPM_CONFIG_STRICT_PEER_DEPS: true
-
-      - name: Setup project for CI run
-        run: npm run setup:ci && npm --prefix admin run setup:ci && npm --prefix site run setup:ci
+          build: "true"
```
johnnyomair: I'd remove the build option and only build in the lint workflow. This mirrors our internal CI/CD setup more closely: first lint, then build.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,41 @@
+name: "Setup"
+description: "Set up Node.js, install dependencies, and verify lockfiles are up to date"
+
+inputs:
+  setup-ci:
+    description: "Run the project's CI setup scripts"
+    default: "true"
```
johnnyomair: This should be the default behavior, no need for an option IMO.

manuelblum: Done — dropped the `setup-ci` input; `setup:ci` now always runs as part of the action.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/lint.yml
manuelblum: Done — removed the `build` input from the action and added an explicit Build step in the lint workflow, after Lint (first lint, then build).

---
_Generated by [Claude Code](https://claude.ai/code)_

johnnyomair (approved):

PR #1656, opened 2026-08-31
Add BFF route example for sending contact form inquiries as mail

VPS-julia:
## Description

Adds an example for a BFF route to the starter: it receives the values of a contact form, validates them and sends them as a mail. The mail is sent in the BFF and not in the API, because the inquiries are of no interest to the CMS.

The route follows the example implementation in [vivid-planet/dextinity#4963](https://github.com/vivid-planet/dextinity/pull/4963) and adds the mail sending in its `try` block. Since the starter contains no form, the validation is reduced to the fields any contact form needs: the address to reply to and the message. A project extends the schema with its own fields.

The corresponding form is **not** part of this PR — the starter only provides the route as an example.

## Changes

- **`site/src/app/[visibility]/[domain]/api/contact-form/route.ts`** (new): `POST` handler. Validates the request body with zod, answers with 400 and the validation error on invalid input, and otherwise sends the message as a mail with nodemailer, with the submitted address in `Reply-To`. Errors are logged and answered with 500. The `[visibility]` and `[domain]` segments are added by the domain rewrite middleware, so a form would submit to `/api/contact-form`.
- **`docker-compose.yml`** and **`.env`**: add Mailpit, which catches all mails sent in development and shows them at http://localhost:8025, together with the `MAIL_*` and `CONTACT_FORM_TO_EMAIL` variables. `MAIL_HOST` and `MAIL_PORT` point at Mailpit; the sender and recipient addresses are left empty, because the starter is boilerplate and a project fills them in itself. The deployment configurations are unchanged — a project supplies its own SMTP credentials there.
- **`site/package.json`**: add `nodemailer`, `zod` and `@types/nodemailer`.
- **`AGENTS.md`**: document the BFF route pattern and the Mailpit service.

## Testing

Sending was tested against Mailpit: the mail arrives with the expected `From`, `To`, `Reply-To` and subject.

---

https://vivid-planet.atlassian.net/browse/COM-2702

https://claude.ai/code/session_014QS28L7gMaUqTqRSJmFejN


VPS-Obi (commented):
I don't think this should be in the Starter as there's no contact form. Instead, the contact form's data should be send as email in the Demo's BFF (it currently only performs validation): https://github.com/vivid-planet/dextinity/blob/main/demo/site/src/app/%5Bvisibility%5D/%5Bdomain%5D/%5Blanguage%5D/api/contact-form/route.tsx#L49

PR #1671, opened 2026-09-04
Generate `.npmrc` from a template on local setup only

VPS-thodax:
## Problem

The root `.npmrc` is committed and symlinked into every package, so `strict-peer-deps=true` and `min-release-age=3` apply in CI as well as locally. The `min-release-age` cooldown is meant to slow developers down on freshly published versions, not to gate CI installs that follow the committed lockfile.

## Solution

- `.npmrc.tpl` is the committed file; `install.sh` copies it to `.npmrc` before the root `npm install`, so every install in the setup sees the settings.
- `setup-project-files.js` links it into the packages only when `CI !== "true"`, so CI runs with no `.npmrc` at all.
- The root `.gitignore` ignores `.npmrc` — the bare pattern covers the root file and every package copy, so the per-package entries are dropped.

## Decisions

- **Starter-only.** The `dextinity` repo is a pnpm workspace with no `.npmrc`; the equivalent settings live in its committed `pnpm-workspace.yaml`.

## Verification

`node setup-project-files.js` links `.npmrc` into all four packages with none of them showing up in `git status -uall`; `CI=true` creates none.

## Further information

- Task: https://vivid-planet.atlassian.net/browse/DEX-3161

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01G4aXwsvMLGJFpCjhUou4W7


VPS-Obi (commented):
I'm hesitant about this. AFAIK `min-release-age` isn't used in CI as we're using `npm ci` instead of `npm install`. This way we loose any other CI-relevant option as well (although I'm unsure if `strict-peer-deps=true` is respected).

Is this really a problem?

